Issue #7, opened 2018-07-08
Module parse failed: Unexpected token (36:9). You may need an appropriate loader to handle this file type.


ghost:
I'm using this in package.json:

```json
...
"cke5-react": "github:ckeditor/ckeditor5-react",
"@ckeditor/ckeditor5-build-classic": "10.1.0",
...
```

in my code this:

```js
...
import CKEditor from "cke5-react/src/ckeditor";
import ClassicEditorBuild from "@ckeditor/ckeditor5-build-classic";
...
```

but it says:

```log
./node_modules/cke5-react/src/ckeditor.js
Module parse failed: Unexpected token (36:9)
You may need an appropriate loader to handle this file type.
|       // Render <div> element which will be replaced by CKEditor.
|       render() {
|               return <div ref={ ref => ( this.domContainer = ref ) }></div>;
|       }
|
```
**Why?**

pomek:
You need to attach JSX loader for the specific component file in your Webpack configuration.

johnunclesam:
I'm using `create-react-app` without ejecting.

johnunclesam:
If I copy the code in a component in my project everything works!

pomek:
Could you install the package using the [`#t/5`](https://github.com/ckeditor/ckeditor5-react/tree/t/5) branch?

It contains a lot of improvements required for the #5. The issue that you described shouldn't occur anymore.

johnunclesam:
Yes it works with that branch. But the problem now is `componentWillReceiveProps( newProps ) {` which freeze my component at startup.

PR #8, opened 2018-07-08
Update ckeditor.js

johnunclesam:
Because of https://github.com/ckeditor/ckeditor5-react/issues/7 I copied all the code from `ckeditor.js` in my project (in a new component file).

I'm using `create-react-app` and eslint. So the errors:

- [eslint] componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops (react/no-deprecated)

- [eslint] shouldComponentUpdate should be placed after UNSAFE_componentWillReceiveProps (react/sort-comp)

- [eslint] shouldComponentUpdate should be placed after componentDidMount (react/sort-comp)

- [eslint] render should be placed after _destroyEditor (react/sort-comp)

- [eslint] Default prop types declarations should be sorted alphabetically (react/jsx-sort-default-props)

- [eslint] Must use destructuring props assignment (react/destructuring-assignment)

**What is still a problem** (and I don't know how to fix):

- [eslint] UNSAFE_componentWillReceiveProps is unsafe for use in async rendering, see https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html (react/no-unsafe)

- Unexpected console statement. (no-console) (62, 9)

- JSX props should not use arrow functions (react/jsx-no-bind) (68, 17)

- Prop type `object` is forbidden (react/forbid-prop-types) (74, 3)

- propType "onChange" is not required, but has no corresponding defaultProp declaration. (react/require-default-props) (77, 3)

- propType "onInit" is not required, but has no corresponding defaultProp declaration. (react/require-default-props) (78, 3)

### DIFF
--- a/src/ckeditor.js
+++ b/src/ckeditor.js
@@ -1,88 +1,87 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React from "react";
+import PropTypes from "prop-types";
 
-export default class CKEditor extends React.Component {
+class CKEditor extends React.Component {
+  constructor(props) {
+    super(props);
 
-	constructor( props ) {
-		super( props );
+    this.editorInstance = null;
+  }
 
-		this.editorInstance = null;
-	}
+  // Update editor data if data property is changed.
+  UNSAFE_componentWillReceiveProps(newProps) {
+    if (this.editorInstance && newProps.data) {
+      this.editorInstance.setData(newProps.data);
+    }
+  }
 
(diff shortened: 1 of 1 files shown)


pomek:
I'm closing this PR because of proposed changes don't match to our guide.

We will introduce a linter and then we will be able to fix errors (if any).

PR #14, opened 2018-07-18
Improvements before publishing

### LINKED ISSUE #1: Initial configuration
Implementation of https://github.com/ckeditor/ckeditor5/issues/1015.

### LINKED ISSUE #3: React 17 will remove `componentWillReceiveProps()` function that is used in component
As in title, the following method won't work with the future version:

https://github.com/ckeditor/ckeditor5-react/blob/346a586fbee5b33a5251506e755474a93f8dcc3d/src/ckeditor.js#L17-L22

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

### LINKED ISSUE #5: Change the name to @ckeditor/ckeditor5-react and publish first dev preview
People started landing here and reporting that this pkg is missing (#2, #4), so let's publish something.

TODO:

* [x] Review the README.
* [x] Write clearly that this is a developers' preview and that we need your feedback.
* [x] Change the name to `@ckeditor/ckeditor5-react` and update the docs.
* [x] https://github.com/ckeditor/ckeditor5-react/issues/5#issuecomment-394964843 - RegExps on Windows
* [x] https://github.com/ckeditor/ckeditor5-react/issues/9 - Check linter
* [ ] Publish!

### LINKED ISSUE #7: Module parse failed: Unexpected token (36:9). You may need an appropriate loader to handle this file type.

### LINKED ISSUE #13: Error when building with create-react-app.
I'm using `create-react-app@1.5.2` and when I build using `npm run build` it says:

```log
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file:

        ./node_modules/@ckeditor/ckeditor5-build-classic/build/ckeditor.js:5:7374

Read more here: http://bit.ly/2tRViJ9

error Command failed with exit code 1.
```

How to do?

Maybe is there a way to disable es5 using just es6 in CRA?

### LINKED ISSUE #15: Add some tests
They will be a great example of how to use the component.

Of course, they also will prevent us before some regression.

pomek:

Other: First developer preview version. It contains ready-to-use `CKEditor` component that allows using CKEditor 5 Builds and CKEditor 5 Framework in React application. Closes #1. Closes #3. Closes #5. Closes #7. Closes #13. Closes #15.


scripts/release.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
pomek: I guess so.

pomek:
@Reinmar, could look on this PR?

ma2ciek:
I'm not sure if it won't be good to add a component with simple usage of the editor component and some tests.

pomek:
I will add some tests but I would like to have a final image of the package.

pomek:
Tests are reported as #15.

src/ckeditor.jsx
```diff
@@ -33,25 +33,25 @@ export default class CKEditor extends React.Component {
...
 				if ( this.props.onInit ) {
-					this.props.onInit( editor );
+					this.props.onInit( this.editor );
 				}
 
 				if ( this.props.onChange ) {
-					const document = this.editorInstance.model.document;
+					const document = this.editor.model.document;
+
 					document.on( 'change', () => {
```
ma2ciek: You can listen to `document.on( 'change:data' )` and remove this `if` below.

src/ckeditor.jsx
```diff
@@ -33,25 +33,25 @@ export default class CKEditor extends React.Component {
...
+				this.editor = editor;
+				this.editor.setData( this.props.data );
 
-				// TODO: Add example using it.
 				if ( this.props.onInit ) {
-					this.props.onInit( editor );
+					this.props.onInit( this.editor );
 				}
 
 				if ( this.props.onChange ) {
```
ma2ciek: I'd move this inside the listener, as this property can be changed at runtime.

README.md
```diff
@@ -0,0 +1,292 @@
...
+
+```bash
+npm install --save @ckeditor/ckeditor5-react @ckeditor/ckeditor5-build-classic
+```
+
+Use CKEditor component inside your project:
+
+```jsx
+import React, { Component } from 'react';
+import './App.css';
```
ma2ciek: You can remove all these `import './App.css';` from the readme.

ma2ciek:
```
 "dependencies": {
    "prop-types": "^15.6.1",	    "prop-types": "^15.6.1",
    "react": "^16.3.2"	    "react": "^16.3.2"
 }
```

I guess, these should be `peerDependencies` and `devDependencies`, not `dependecies`, so you won't install two versions of React installing that package.

pomek:
According to https://github.com/facebook/prop-types#how-to-depend-on-this-package, `prop-types` will be marked as a dependency and `react` as a peer-dependency.

pomek:
IMO it's ready for review once again and merge if eveyrthing will be fine.

pomek:
An issue with `npm run build` (in fact you could not build the app) was resolved.

dist/ckeditor.js
```diff
@@ -0,0 +1,6 @@
+/*!
```
ma2ciek: Should we provide `dist` directory in the github repo?

webpack.config.js
```diff
@@ -0,0 +1,67 @@
...
+						// Preserve CKEditor 5 license comments.
+						comments: /^!/
+					}
+				}
+			} )
+		]
+	},
+
+	plugins: [
+		new webpack.BannerPlugin( {
```
ma2ciek: IMO we shouldn't do that. That build contains the whole react library.

ma2ciek: And thus I'd add React to the extrnals to `webpack.config.js` when building our react component.

dist/ckeditor.js
pomek: Why not? https://github.com/ckeditor/ckeditor5-build-classic/tree/master/build

webpack.config.js
ma2ciek: It'll also prevent loading React few times.

dist/ckeditor.js
ma2ciek: So you'll be updating the `dist` directory every CKE5 release? 😄

pomek: For what? If the source will change, the dist will be updated. CKE5 releases do not touch the package.

package.json
```diff
@@ -38,20 +38,21 @@
     "karma-webpack": "^3.0.0",
     "minimist": "^1.2.0",
     "mocha": "^5.2.0",
-    "prop-types": "^15.6.1",
-    "react": "^16.3.2",
     "react-dom": "^16.4.1",
     "sinon": "^6.1.3",
     "webpack": "^4.2.0"
```
ma2ciek: You can update to `4.16`. And the `webpack-cli` is missing.

pomek: I don't think that webpack-cli is required.

webpack.config.js
pomek: Done.

karma.conf.js
```diff
@@ -0,0 +1,266 @@
...
+			'source-map': false,
+		}
+	};
+
+	const options = minimist( args, config );
+
+	options.sourceMap = options[ 'source-map' ];
+	options.browsers = options.browsers.split( ',' );
+
+	// Delete all aliases because we don't want to use them in the code.
```
ma2ciek: ?

tests/ckeditor.jsx
```diff
@@ -0,0 +1,268 @@
...
+
+		sandbox.restore();
+	} );
+
+	it( 'should call "Editor#create()" method during initialization the component', () => {
+		sandbox.stub( Editor, 'create' ).resolves( new Editor() );
+
+		wrapper = mount( <CKEditor editor={ Editor } /> );
+
+		expect( Editor.create.calledOnce ).to.be.true;
```
ma2ciek: You can replace chai's expects with sinon asserts if it touches spies / mocks.

karma.conf.js
pomek: ?

webpack.config.js
```diff
@@ -0,0 +1,68 @@
...
+const webpack = require( 'webpack' );
+const { bundler } = require( '@ckeditor/ckeditor5-dev-utils' );
+const UglifyJsWebpackPlugin = require( 'uglifyjs-webpack-plugin' );
+
+module.exports = {
+	context: __dirname,
+
+	devtool: 'source-map',
+	performance: { hints: false },
+	externals: { react: 'React' },
```
ma2ciek: We should pass an object here to satisfy all environments: https://webpack.js.org/configuration/externals/#externals

src/ckeditor.jsx
```diff
@@ -1,23 +1,23 @@
...
+		// @see: https://docs.ckeditor.com/ckeditor5/latest/api/module_core_editor_editor-Editor.html
+		this.editor = null;
 	}
 
-	// Update editor data if data property is changed.
-	componentWillReceiveProps( newProps ) {
-		if ( this.editorInstance && newProps.data ) {
-			this.editorInstance.setData( newProps.data );
+	componentDidUpdate( newProps ) {
+		if ( this.editor && newProps.data ) {
```
ma2ciek: I'm thinking about cases when someone stores editor content in the external place.

Then it'd look like this:

editor changes => ckeditor component changes => parent component saves the data => some redux store detects changes => redux store updates => component updates with same props => editor changes. And it'd create an infinite loop for now.

I'd write it like this:
```js
componentDidUpdate() {
	if ( this.editor && this.editor.getData() !== this.props.data ) {
		this.editor.setData( this.props.data );
	}
}
```

It should be safer.

karma.conf.js
ma2ciek: Ok. Didn't know that aliases are the part of minimist's results.

webpack.config.js
ma2ciek: The build lands in the `dist` directory, which you will publish on the NPM. And then you can add it using e.g. `<script>` tag or load it via `import CKEditor from '@ckeditor/ckeditor5-react';`. So only we build this package, but it can be used in different ways.

pomek: Sorry, I don't understand. 

There are two ways to use it and both ways are tested: (1) in `tests/ckeditor-classiceditor.jsx` we check the build and (2) the sample checks whether an editor works.

src/ckeditor.jsx
ma2ciek: React + redux is pretty common. And people would need to hack to prevent component updates.

> Except that, this.editor.getData() never returns an empty string. If this.props.data will be empty, this if will always return true.

Right, but once the props change to `<p>&nbsp;</p>` the cycle ends, because the `props.data` and `editor.getData()` will return the same string.

pomek: > React + redux is pretty common. And people would need to hack to prevent component updates.

Would be awesome to see almost one ticket which explains the problem. In fact, I'm making my first steps with React and I would like more opinions. At this moment, people want to use the package and they cannot. That's the goal of the PR. Publish and allows using (and testing). If something will be broken, we will fix that. Now, I cannot imagine the whole picture.

> Right, but once the props change to <p>&nbsp;</p> the cycle ends, because the props.data and editor.getData() will return the same string.

Sure? Are the properties bound two-ways? Component => CKEditor will update the content, but content updated by a user will not change the Component property. Maybe I'm wrong. I don't insist.

PR #45, opened 2018-10-25
Fixed invalid build name for builds on BrowserStack.

### LINKED ISSUE #44: Add current timestamp to project name for builds on BrowserStack
ATM, every build has the same name: `ckeditor5 react`. Recently we upgraded a name of builds by adding a current timestamp as a suffix. This patch should be applied to this repo as well (https://github.com/ckeditor/ckeditor5-dev/pull/432).

cc: @oleq, @ma2ciek

pomek:

Internal: Add current timestamp to project name for builds on BrowserStack. Thanks to that, badges will always show a status of the latest build and its link open the build instead of the first created build. Closes #44.

### DIFF
--- a/scripts/utils/getkarmaconfig.js
+++ b/scripts/utils/getkarmaconfig.js
@@ -100,7 +100,7 @@ module.exports = function getKarmaConfig() {
 		karmaConfig.browserStack = {
 			username: process.env.BROWSER_STACK_USERNAME,
 			accessKey: process.env.BROWSER_STACK_ACCESS_KEY,
-			build: process.env.TRAVIS_REPO_SLUG,
+			build: getBuildName(),
 			project: 'ckeditor5'
 		};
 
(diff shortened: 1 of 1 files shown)


pomek:
image

It just works! :tada:

pomek:
> Can't we expose this util in one of our packages?

We could but all integrations have the same code. It means that the issue touches 3 integrations and requires more changes now.

Reinmar:
OK, so in the other 2 integrations this piece of code is already in use? That indeed means that it does not make sense to touch this now. However, if we'll see that we need to update this code in all 3 frameworks one day, then let's move it to one place then.

pomek:
> OK, so in the other 2 integrations this piece of code is already in use?

Yes, they do.

> However, if we'll see that we need to update this code in all 3 frameworks one day, then let's move it to one place then.

I will remember.

PR #46, opened 2018-11-01
Update README.md

zelongc:
listing Props in the README could help newbies (like me) save a lot of time from googling around how to config a ck-editor5 react component.

Type: Message. Closes #000.

---

### Additional information

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,18 @@ After cloning this repository, install necessary dependencies:
 npm install
 ```
 
+## Props
+
+```
+CKEditor.propTypes = {
+	editor: PropTypes.func.isRequired,
+	data: PropTypes.string,
+	config: PropTypes.object, // Config object comes here.
+	onChange: PropTypes.func,
+	onInit: PropTypes.func
+};
+```
+
 ### Executing tests
 
 ```bash


pomek:
Hi @zelongc,

Thanks for the PR. However, all properties are listed in the documentation (https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/react.html#component-properties). The link to this page is located at the beginning of the README. 

I don't guess that repeating the same content will help.

zelongc:
My concern was, there could be a lot of non-english speakers who can be eassily frightened by a screen of english characters just like me.. Providing a list of props in README doesn't hurt anyone and could potentially save them a lot of time.

Thanks for the review @pomek and your awesome work.

Reinmar:
The question is whether this component can really be used without reading the docs. There are a couple of things one needs to know now to use it.

Also, we should be cautious to not make a false impression that this is the entire documentation here in the `README.md`. The more we'll add here, the bigger the chance that people will miss all the links to the actual documentation.

Reinmar:
PS. That said – I believe we should improve the "Documentation" section because it's not very visible now either. @ma2ciek had the idea to put there the table of contents from https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/react.html. What do you think about it?

PR #52, opened 2018-11-21
"onBlur" and "onFocus" properties

### LINKED ISSUE #49: Allow other events (e.g. onBlur) on editable container
In our application, we validate form fields after the user has entered data by hooking onto the `onBlur` event. However, this behavior doesn't work with the current implementation of the CKEditor React component.

It'd be nice if we could pass extra props (`data-*` attributes, event handlers, etc) to the editor component and have them rendered along with the existing properties.

pomek:

Feature: Introduced `onFocus` and `onBlur` properties.  Closes #49.

---

### Additional information

Improvements in the component's documentation – https://github.com/ckeditor/ckeditor5/pull/1364.


tests/_utils-tests/editor.js
```diff
@@ -0,0 +1,56 @@
+/**
```
ma2ciek: Should we create tests for mocks?

tests/_utils/editor.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+/**
+ * Mock of class that representing a basic, generic editor.
+ *
+ * @see: https://ckeditor.com/docs/ckeditor5/latest/api/module_core_editor_editor-Editor.html
+ */
+export default class Editor {
+	constructor() {
+		this.model = Editor._model;
```
ma2ciek: why it can't be `this.model = { document: createDocument() }`?

tests/_utils-tests/editor.js
pomek: Why not? This test ensures API of the Editor mock which is used in other tests.

tests/_utils/editor.js
pomek: https://github.com/ckeditor/ckeditor5-react/pull/52/files/f1ceee7fb77036c982de21b8be171fb1573237e6#diff-178f392effc12f072e31bd8a383f0a50R31

If you mock those properties after initialization the component, call those functions: `modelDoc.on` and `viewDoc.on` won't be detected.

ma2ciek:
It would be useful to add these events to the sample. Also, the sample should have an infobox to open the console.

pomek:
Done.

PR #55, opened 2018-11-26
Prevent to an infinite loop when [data] property is missing

### LINKED ISSUE #39: Check props.data in the componentDidUpdate() lifecycle
Follow-up of the https://github.com/ckeditor/ckeditor5-react/issues/37 and https://github.com/ckeditor/ckeditor5-react/issues/37#issuecomment-422707443

https://github.com/ckeditor/ckeditor5-react/blob/383b25d35beb54788f91185ea6c30615b56bb2b6/src/ckeditor.jsx#L18-L22

As @pomek proposed, we should check if the `this.props.data` is equal to `undefined` to prevent an infinite update loop.

pomek:

Fix: Prevent to an infinite loop when the `[data]` property is missing. Closes #39.


src/ckeditor.jsx
```diff
@@ -15,8 +15,20 @@ export default class CKEditor extends React.Component {
 		this.editor = null;
 	}
 
+	shouldComponentUpdate( nextPros ) {
+		if ( 'data' in nextPros ) {
```
ma2ciek: It should be inside the componentDidUpdate part IMO. Now if somebody would like to change other prop and don't provide the `data` the component won't be updated.

WDYT?

pomek: I need to take a deep breath… I will try to write another test case which will reproduce mentioned issue and we will see what happen.

ma2ciek: I've checked it and it somehow works with it. It's because the whole logic is in the `componentDidMount lifecycle. So the component is rendered once and its props are being updated anyway. What the  `shouldComponentUpdate` actually does is preventing calling `componentDidMount` and `render` methods. But in our case `render` method does actually nothing and preventing calling `componentDidMount` is the only visible effect.

But anyway, the logic now is split into two methods while it can be in one place.

pomek: But right now, we don't have other properties that must be updated.

Events for props: `onChange`, `onFocus` and `onBlur` are always executed. But the `onChange` callback will be invoked only if `this.props.onChange` is a function.

PR #56, opened 2018-11-27
Support for read-only mode

### LINKED ISSUE #53: Introduce [disabled] property
Both integrations (Angular and Vue) allows changing the editor to read-only mode.

In React you can do it by saving the `editor` from `onInit()` callback. If we will introduce the `[disabled]` property, switching between those modes should be easier.

pomek:

Feature: Introduced the `[disabled]` property which allows switching the editor to read-only mode. Closes #53.


tests/ckeditor.jsx
```diff
@@ -372,6 +372,64 @@ describe( 'CKEditor Component', () => {
 				} );
 			} );
 		} );
+
+		describe.only( '#disabled', () => {
```
pomek: ❤️

pomek:
Because https://github.com/ckeditor/ckeditor5/pull/1364 is still open, I added the new property in mentioned PR.

src/ckeditor.jsx
```diff
@@ -103,11 +127,13 @@ CKEditor.propTypes = {
 	onChange: PropTypes.func,
 	onInit: PropTypes.func,
 	onFocus: PropTypes.func,
-	onBlur: PropTypes.func
+	onBlur: PropTypes.func,
+	disabled: PropTypes.oneOfType( [ PropTypes.bool, PropTypes.string ] )
```
ma2ciek: I don't agree here. Why we should support multiple types here? Can't we just support only `boolean`?

pomek: Why not? You can do `disabled="true"` (as I did before I discovered `disabled={true}`) and it still will work.

ma2ciek: But `disabled="true"` isn't the preferred way to pass a boolean option to react component as you find out yourself... IMO we should support more than one type if there're a concept and more logic behind the other types (like string and array of strings).

pomek: Now the component will accept boolean value only.

sample/index.html
```diff
@@ -51,6 +56,16 @@ <h2>Sample</h2>
 			} ),
 			document.getElementById( 'root' )
 		);
+
+		document.getElementById( 'readOnly' ).addEventListener( 'click', function() {
+			window.editor.isReadOnly = !window.editor.isReadOnly;
```
ma2ciek: What if the editor isn't available here? 😄

pomek: I moved this code to `onInit` component's callback.

PR #69, opened 2019-02-27
Changed a way how an initial data are passed to the editor

### LINKED ISSUE #68: Stop using `editor.setData()` after editor's initialization
It causes the content of the initialized editor is being changed. 

There is no problem if only a single user edits the content. But what if two and more users want to edit the editor. Every new connection will reset the editor to the initial state. The entire history of the document will disappear.

We need to find another way to set initialized data.

pomek:

Fix: Changed a way how an initial data are passed to an editor instance. Previously it used the `editor.setData()` method which produces invalid results in collaboration. Now the data are injected to the container, where the editable will be enabled. Closes #68.


pjasiun:
> This change fixes the problem indeed, however, I'm not sure if allowing `innerHTML` is the best we can do here.

Unfortunately, at this stage, it might be the best what we can do. We need to work on the editor initialization process, but at this moment I have no better ideas than `innerHTML` (we do such thing in our samples).

PR #79, opened 2019-03-21
The component will not be updated by React itself

### LINKED ISSUE #75: InlineEditor OnChange problem
Hello 

İf I use InlineEditor in React. OnChange event just once update state and then has error in console.

My Code:
```
import React, { Component } from 'react';
import logo from './logo.svg';
import './App.css';

import CKEditor from '@ckeditor/ckeditor5-react';
import ClassicEditor from '@ckeditor/ckeditor5-build-classic';
import InlineEditor from '@ckeditor/ckeditor5-build-inline';

class App extends Component {
  constructor(props) {
    super(props);

    this.state = {
      content: 'Hello'
    }
  }

  handleEditorChange() {
    return (event, editor) => {
      this.setState({ content: editor.getData() });
      console.log(this.state);
    }
  }

  render() {
    return (
      <div className="App">
        <div>
          <CKEditor
                      editor={ InlineEditor }
                      data={this.state.content}
                      onInit={ editor => {
                          // You can store the "editor" and use when it is needed.
                          console.log( 'Editor is ready to use!', editor );
                      } }
                      onChange={ this.handleEditorChange()}
                      onBlur={ editor => {
                          console.log( 'Blur.', editor );
                      } }
                      onFocus={ editor => {
                          console.log( 'Focus.', editor );
                      } }
                  />

        </div>
      </div>
    );
  }
}

export default App;
```

and console error:
image

### LINKED ISSUE #78: Editor freezes when [data] is string
Not related to reported issue but if you pass as `[data]` any string, after changing editor's content, you will freeze the app. I guess, we should remove this part of the code: https://github.com/ckeditor/ckeditor5-react/blob/a31de76118130eae1ac8ae82c5c02917f996e211/src/ckeditor.jsx#L25-L27

We shouldn't care about updating editor's content if the state has changed. A user can do it manually just by calling the callback as the second parameter in `setState` function.

_Originally posted by @pomek in https://github.com/ckeditor/ckeditor5-react/issues/75#issuecomment-475248189_

pomek:

Fix: The `<CKEditor>` component will not be updated by React itself. The editor won't freeze if `#data` property was specified as a static string. Closes #75. Closes #78.

### DIFF
--- a/src/ckeditor.jsx
+++ b/src/ckeditor.jsx
@@ -19,6 +19,10 @@ export default class CKEditor extends React.Component {
 
 	// This component should never be updated by React itself.
 	shouldComponentUpdate( nextProps ) {
+		if ( this._shouldUpdateContent( nextProps ) ) {
+			this.editor.setData( nextProps.data );
+		}
+
 		if ( 'disabled' in nextProps ) {
 			this.editor.isReadOnly = nextProps.disabled;
 		}
@@ -96,6 +100,23 @@ export default class CKEditor extends React.Component {
 					this.editor = null;
 				} );
 		}
+	}
+
+	_shouldUpdateContent( nextProps ) {
+		// Check whether `nextProps.data` is equal to `this.props.data` is required if somebody defined the `#data`
+		// property as a static string and updated a state of component when the editor's content has been changed.
+		// If we avoid checking those properties, the editor's content will back to the initial value because
+		// the state has been changed and React will call this method.
+		if ( this.props.data === nextProps.data ) {
+			return false;
+		}
+
+		// We should not change data if the editor's content is equal to the `#data` property.
+		if ( this.editor.getData() === nextProps.data ) {
+			return false;
+		}
+
+		return true;
 	}
 }
 
--- a/src/ckeditor.jsx
+++ b/src/ckeditor.jsx
@@ -17,18 +17,13 @@ export default class CKEditor extends React.Component {
 		this.domContainer = React.createRef();
 	}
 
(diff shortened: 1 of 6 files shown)


pomek:
I have an idea of how to update the editor's content when the state has changed. Let me introduce some improvements to this PR.

pomek:
@ma2ciek, go ahead.

ma2ciek:
I'm not glad that the editor state changes will go to the `shouldComponentUpdate`, but since for now it's the only possible way to prevent React from removing CKEditor5 DOM elements on rerendering, I'm merging it. Maybe we'll find a better approach in the future.

src/ckeditor.jsx
```diff
@@ -17,18 +17,17 @@ export default class CKEditor extends React.Component {
 		this.domContainer = React.createRef();
 	}
 
-	componentDidUpdate() {
-		if ( !this.editor ) {
-			return;
+	// This component should never be updated by React itself.
+	shouldComponentUpdate( nextProps ) {
+		if ( this._shouldUpdateContent( nextProps ) ) {
+			this.editor.setData( nextProps.data );
```
faizaankhan, line 23: This was not the right place to do this. :|

faizaankhan: I have editor content saved as a draft in local storage, and it is picked up and sent to CKEditor as a prop in data. Of course the value changes for data, but CKeditor doesn't re-render.

pomek: Please, report an issue if you think that this part of the code is invalid.

PR #85, opened 2019-03-28
Component will not update anything if it is not ready

### LINKED ISSUE #83: TypeError: Cannot read property 'insertBefore' of null and Cannot set property 'isReadOnly' of null
Hi guys, I'm trying to use your CKEditor 5 React component in a Create React App v2, to build a big form, together with other standard HTML form inputs, and other instances of CKEditor all in the same form.
The form itself is driven by a HOC with a data structure which drives all of the form state, something in the lines of:
```
state = {
  resource: {
    name: {
      value: '',
      name: 'name',
      type: 'text',
    },
    content: {
      value: '',
      name: 'content',
      type: 'ckeditor',
    }
    parent_id: {
      value: '',
      name: 'parent_id',
      type: 'select',
      // This gets populating by data-fetching from an external REST API
      values: [],
    }
  }
}
```
Part of this state/schema waits for additional values to be updated after a fetch call is fired to a REST API, basically in order to populate select options and values.
This causes a few setStates right at the moment of fetch (to update UI state with spinners), and at the moment of receiving back those values (simply updating the state with them).
While this works seamlessly with other form inputs, it looks like CKEditor throws an error if there are too many setStates close to each other, it looks like the this.editor instance in the CKEditor component is not refreshed in time and a ```TypeError: Cannot read property 'insertBefore' of null``` is thrown.

I've created a CodeSandbox for you to try, while it is not exactly the same code I'm using (the CodeSandbox leverages hooks, and setInterval for firing off quick and close to each other setState calls), it can hopefully showcase the problem.
https://codesandbox.io/s/py52ll3v4j?fontsize=14

Additional to that (this I can't really reproduce on CodeSandbox unfortunately), CKEditor throws another error on my CRA when the disabled prop is passed to the CKEditor component and these repeated setState calls throw the error ```TypeError: Cannot set property 'isReadOnly' of null```
Which seems to occur since the latest v1.1.2 (was not occurring on v1.1.1) due to [this edit](https://github.com/ckeditor/ckeditor5-react/commit/038033de1a35878c41f6350865c8641b27291bf6) in shouldComponentUpdate.

Could you point me in the right direction if I'm doing something wrong in the way I'm using the component or if there's an issue with CKEditor itself?

Thanks guys,
Marco

pomek:

Fix: The `<CKEditor>` component will not update anything if it is not ready. Closes #83.

### DIFF
--- a/src/ckeditor.jsx
+++ b/src/ckeditor.jsx
@@ -18,6 +18,10 @@ export default class CKEditor extends React.Component {
 
 	// This component should never be updated by React itself.
 	shouldComponentUpdate( nextProps ) {
+		if ( !this.editor ) {
+			return false;
+		}
+
 		if ( this._shouldUpdateContent( nextProps ) ) {
 			this.editor.setData( nextProps.data );
 		}
--- a/tests/ckeditor.jsx
+++ b/tests/ckeditor.jsx
@@ -123,6 +123,24 @@ describe( 'CKEditor Component', () => {
 	} );
 
 	describe( 'properties', () => {
+		// See: #83.
+		it( 'does not update anything if component is not ready', () => {
+			const editorInstance = new Editor();
+
+			sandbox.stub( Editor, 'create' ).resolves( editorInstance );
+
+			wrapper = mount( <CKEditor editor={ Editor } /> );
+
+			const component = wrapper.instance();
+			let shouldComponentUpdate;
+
+			expect( () => {
+				shouldComponentUpdate = component.shouldComponentUpdate( { disabled: true } );
+			} ).to.not.throw();
+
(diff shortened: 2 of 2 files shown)


pomek:
_Note: After merging the PR, write a comment under #83 that the second issue should be reported once again if an author will be able to reproduce it in some testing sample._

src/ckeditor.jsx
```diff
@@ -18,6 +18,10 @@ export default class CKEditor extends React.Component {
 
 	// This component should never be updated by React itself.
 	shouldComponentUpdate( nextProps ) {
+		if ( !this.editor ) {
```
ma2ciek: Seems like a good check 😄

PR #127, opened 2019-12-20
Add optional onError callback

### LINKED ISSUE #123: Could we expose an optional callback handler to be called here instead of only logging errors?
https://github.com/ckeditor/ckeditor5-react/blob/9c60780b8069392a2db346dc872660c2aabf51fa/src/ckeditor.jsx#L95

ansorensen:

Feature: Introduced support for `onError` callback that is being called if an error occurred during the editor's initialization. Closes #123.

---

### Additional information

Because CKEditor5 does not expose an onError listener in events, for right now all we can expose is a callback for errors thrown when CKEditor5 is initialized in the CKEditor5-React component.

### DIFF
--- a/src/ckeditor.jsx
+++ b/src/ckeditor.jsx
@@ -92,7 +92,12 @@ export default class CKEditor extends React.Component {
 				} );
 			} )
 			.catch( error => {
-				console.error( error );
+				if (this.props.onError) {
+					this.props.onError( error )
+				}
+				else {
+					console.error( error );
+				}
 			} );
 	}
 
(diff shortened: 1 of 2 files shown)


pomek:
@ansorensen, sorry for the late response. I would like to say thank you for the PR. It has been merged. I hope the new version of the package will be released this week.

pomek:
Released as [v2.1.0](https://github.com/ckeditor/ckeditor5-react/releases/tag/v2.1.0).

ma2ciek:
TBH I'm not sure if it not collides with the planned feature: [https://github.com/ckeditor/ckeditor5/issues/4693](https://github.com/ckeditor/ckeditor5/issues/4693) :smile:

ansorensen:
The Watchdog feature does pretty much cover this use case. So if ckeditor5-react used Watchdog and offered optional callbacks for Watchdog methods like 'onError' and 'onRestart' that could easily replace the current onError implementation.

PR #134, opened 2020-02-06
update webpack.config to replace UglifyJS with Terser

### LINKED ISSUE #133: uglifyjs-webpack-plugin vulnerability
In the `webpack.config`, the minimizer plugin used currently is `UglifyJsWebpackPlugin`. This Plugin has a dependency `serialize-javascript` which has a cross-site scriting vulnerability: https://npmjs.com/advisories/1426. 

Uglifyjs-webpack-plugin itself is archived so the webpack config should switch to to use `terser-webpack-plugin`.

tx-michael:
Closes #133 

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
     "sinon": "^7.2.3",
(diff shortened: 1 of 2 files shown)


Reinmar:
cc @pomek

PR #178, opened 2020-09-08
Add support for the Context and Watchdog features

### LINKED ISSUE #118: Support for Context and Watchdog features
Hi there, how do you recommend to add Watchdog to this ? Thanks!

### LINKED ISSUE #168: onChangeCalled when ck-editor should have been removed
## 📝 Provide detailed reproduction steps (if any)

In [this sandbox](https://codesandbox.io/s/ckeditor-5-arrays-removing-i629j?file=/src/index.js), I create an array of ckeditor-react instances. The value of these editors can be changed. Or the entire value the editor represents can be removed from the array.

1. Click the 'x' button by the second editor

### ✔️ Expected result
The 2nd line (editor with 'two' and its button) should be removed. Expect only the handle remove function to called.

### ❌ Actual result
There are still 3 editors, now display "one", "three", "three". You can see from the console logs that the new state _should_ be ["one", "three"]. Instead, the onChange is fired and resolve _after_ the removal, resulting in ["one", "\<p>three\</p>", "three"]

## 📃 Other details

* Browser: chome
* OS: mac
* CKEditor version: 2.1.0
* Installed CKEditor plugins: out of the box classic build (20.0.0)

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

### LINKED ISSUE #169: Multiple editor instances deletion triggers onChange bug in React
## 📝 Provide detailed reproduction steps (if any)
Link to Codesandbox with the issue recreated - [https://codesandbox.io/s/ckeditor5-onchange-bug-e0nt4?file=/src/App.js](https://codesandbox.io/s/ckeditor5-onchange-bug-e0nt4?file=/src/App.js)

### ✔️ Expected result

When I delete the second item in the list, the first item and third item still remain.

### ❌ Actual result

When I delete the second item in the list, the third item duplicates and takes up index 1 and 2.

## 📃 Other details

When I use the onBlur event instead of onChange, it works as expected.

* Browser: Chrome 
* OS: …
* CKEditor version: 5
* Installed CKEditor plugins: Default built in plugins - 	Alignment, Autoformat, BlockQuote, Bold, CKFinder, CKFinderUploadAdapter, Essentials, FontColor, FontSize, Heading, Highlight, HorizontalLine, Image, ImageCaption, ImageStyle, ImageToolbar, ImageUpload, Indent, IndentBlock, Italic, Link, List, MediaEmbed, Paragraph, PasteFromOffice, RemoveFormat, Table, TableToolbar, TextTransformation, Underline

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

ma2ciek:

Feature: The `<CKEditor>` component contains the built-in [watchdog](https://ckeditor.com/docs/ckeditor5/latest/features/watchdog.html) feature. Closes #118. 

Feature: Introduced the `<CKEditorContext>` component that supports the [context](https://ckeditor.com/docs/ckeditor5/latest/features/collaboration/context-and-collaboration-features.html) feature.

Feature: Added the `id` property which is used to distinguish different documents. When this property changes, the component restarts the underlying editor instead of setting data on it, which allows e.g. for switching between collaboration documents and fixes a couple of issues (e.g. the `onChange` event no longer fires during changing the document). Closes #168. Closes #169.

Feature: The `onError()` callback will be called with two arguments. The first one will be an error object (as it was before the release 3+). A second argument is an object that contains two properties:

  * `{String} phase`: `'initialization'|'runtime'` - Informs when the error has occurred (during the editor/context initialization or after the initialization).
  * `{Boolean} willEditorRestart` - When `true`, it means that the editor component will restart itself.
  * `{Boolean} willContextRestart`- When `true`, it means that the context component will restart itself.

    The `willEditorRestart` property will not appear when the error has occurred in the context feature. 
    The `willContextRestart` property will not appear when the error has occurred in the editor.

---

* _Add to the release summary:_ Both components (`<CKEditor>` and `<CKEditorContext>`) will internally use the [`Watchdog`](https://ckeditor.com/docs/ckeditor5/latest/api/module_watchdog_watchdog-Watchdog.html) class that restarts the [editor](https://ckeditor.com/docs/ckeditor5/latest/api/module_watchdog_editorwatchdog-EditorWatchdog.html) or [context](https://ckeditor.com/docs/ckeditor5/latest/api/module_watchdog_contextwatchdog-ContextWatchdog.html) when an error occurs.

* _Add to the release summary:_ The API of the entry point of the package has changed.

The previous way how the package was imported:

```js
import CKEditor from '@ckeditor/ckeditor5-react';
```

will not work with the release (3+). Use:

```js
import { CKEditor } from '@ckeditor/ckeditor5-react';
```

---

BREAKING CHANGE: The [entry point](https://github.com/ckeditor/ckeditor5-react/blob/master/src/index.js) of the package has changed. The default import was removed since the package provides more than a single component now. Use `import { CKEditor } from '@ckeditor/ckeditor5-react'` instead of `import CKEditor from '@ckeditor/ckeditor5-react';`.

BREAKING CHANGE: The `onInit` property was renamed to `onReady` and can be called multiple times (after the initialization and after the component is ready when an error occurred).

---

### Additional information

[TODO]: Add a link to PR in ckeditor5.


ma2ciek:
I see there're some uncovered lines touching the error handling mechanisms. I'll cover them later today.

README.md
```diff
@@ -57,6 +57,12 @@ Build a minified version of the package that is ready to publish:
 npm run build
 ```
 
+### Running samples
+
+All samples are available in the `samples/` directory. They can be open in the browser to check if the component works in some real-world scenarios.
+
+The `context.html` requires build that can be taken from https://github.com/ckeditor/ckeditor5-collaboration-samples/blob/master/comments-outside-of-editor/build/cksource.js or another build that exports both, the `ClassicEditor` and `Context`. Also, there're some prompts that need to be filled during the first open.
```
pomek, line 64: Couldn't we import the script directly in the HTML sample?

```html
<script src="https://raw.githubusercontent.com/ckeditor/ckeditor5-collaboration-samples/master/comments-outside-of-editor/build/cksource.js"></script>
```

src/ckeditor.jsx
```diff
@@ -5,15 +5,36 @@
...
+
+		if ( this.props.contextWatchdog ) {
+			try {
+				return this.props.contextWatchdog.getItem( this._id );
+			} catch ( err ) {
+				return null;
+			}
+		}
+
+		return null;
```
pomek, line 37: The question is – how to get an editor instance when a user does not use Watchdog and Context features?

pomek: Ok, I see that Watchdog is now built-in in the component.

src/context.jsx
```diff
@@ -0,0 +1,75 @@
...
+						} );
+					}
+
+					return child;
+				} ) }
+			</React.Fragment>
+		);
+	}
+
+	_isChildAnEditorComponent( child ) {
```
pomek, line 33: `s/_isChildAnEditorComponent/_isChildEditorComponent`

src/context.jsx
```diff
@@ -0,0 +1,75 @@
...
+		// Use duck typing as the CKEditor component can't be cross-imported.
+		return (
+			typeof child.props.editor === 'function'
+		);
```
pomek, line 37:
```suggestion
		return typeof child.props.editor === 'function';
```

src/utils/uid.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+// A copy of @ckeditor/ckeditor5-utils/src/uid.js
```
pomek, line 6: Let's add the dependency in `package.json` and use it instead of copying the util.

tests/ckeditor-integration.jsx
```diff
@@ -69,9 +82,8 @@ describe( 'CKEditor Component - integration', () => {
 				document.body.appendChild( div );
 
 				return new Promise( resolve => {
-					component = ReactDOM.render( <AppUsingState />, div ); // eslint-disable-line react/no-render-return-value
-
-					setTimeout( resolve );
+					// eslint-disable react/no-render-return-value
```
pomek, line 85:
```suggestion
					// eslint-disable-next-line react/no-render-return-value
```

tests/context.jsx
```diff
@@ -0,0 +1,100 @@
...
+	} );
+
+	describe( 'initialization', () => {
+		it( 'should create an instance of the ContextWatchdog', async () => {
+			wrapper = mount( <Context context={ CKEditorContextMock } /> );
+
+			const component = wrapper.instance();
+
+			expect( component.contextWatchdog ).to.be.an( 'object' );
+		} );
```
pomek, line 46: In fact, you don't check whether `component.contextWatchdog` is an instance of `ContextWatchdog` (`@ckeditor/ckeditor5-watchdog/src/contextwatchdog`).

tests/context.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+		it( 'should render inner elements', async () => {
```
pomek, line 48:
```suggestion
		it( 'should render its children', async () => {
```

tests/context.jsx
```diff
@@ -0,0 +1,100 @@
...
+					<CKEditor editor={EditorMock}></CKEditor>
+				</Context>
+			);
+
+			const component = wrapper.instance();
+
+			expect( component.contextWatchdog ).to.be.an( 'object' );
+			expect( wrapper.childAt( 0 ).name() ).to.equal( 'CKEditor' );
+			expect( wrapper.childAt( 0 ).prop( 'contextWatchdog' ) ).to.be.an( 'object' );
+		} );
```
pomek, line 72: Missing tests:

- scenarios with initial data – does it render correctly?
- scenario with specified the configuration in the `CKEditor` component – will it be merged with context? 
- `Context` with multiple editors (with the same and with different configurations)

Also, I would rename the test case: `should pass the context watchdog to children editor component`

tests/context.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+		it( 'should render the inner editor component', async () => {
```
pomek, line 74:
```suggestion
		it( 'should render the children editor component', async () => {
```

tests/context.jsx
```diff
@@ -0,0 +1,100 @@
...
+			const editorCreateSpy = sinon.spy( EditorMock, 'create' );
+
+			await new Promise( ( res, rej ) => {
+				wrapper = mount(
+					<Context context={ CKEditorContextMock } onError={ rej } >
+						<CKEditor editor={EditorMock} onInit={ res } onError={ rej } />
+					</Context>
```
pomek, line 81: Which `rej` should be called? `Context#onError` or `CKEditor#onError`?

Also, would be good to have an opposite test. If the `Context` component failed here, we should also check what will happen if `CKEditor` fails.

tests/ckeditor.jsx
```diff
@@ -464,67 +444,138 @@ describe( 'CKEditor Component', () => {
...
-					done();
-				} );
+			await new Promise( ( res, rej ) => {
+				onInitCallback = res;
+				onErrorCallback = rej;
+
+				watchdog._restart();
 			} );
+
+			sinon.assert.calledTwice( onInitSpy );
```
pomek, line 563: I believe `onInit` should be called once. From the user point of view, the Watchdog restart mechanism is transparent. The user does not know that the editor failed and it was restarted.

I would be surprised that the `onInit()` hook was called more than once.

tests/ckeditor.jsx
```diff
@@ -464,67 +444,138 @@ describe( 'CKEditor Component', () => {
...
+
+			expect( editor1 ).to.be.instanceOf( Editor );
+			expect( editor2 ).to.be.instanceOf( Editor );
+
+			expect( editor1 ).to.not.equal( editor2 );
 		} );
 	} );
 } );
+
+describe( 'CKEditor.Context', () => {
```
pomek, line 577: There should be also a test that checks whether the component created by using `CKEditor.Context` works since users can also use this way to use the feature.

pomek (changes requested):
First-round. Mostly I am missing tests.

tests/context.jsx
ma2ciek: That's a good point :smile:

tests/ckeditor.jsx
ma2ciek: I don't think so. If this is just the same thing as already tested I think that we can just ensure that the correct thing is exported.

tests/ckeditor.jsx
ma2ciek: Maybe the method name isn't accurate enough. In other components, there's a `ready` event.

But the second call of such event/callback is necessary in order to attach some listeners to the newly created editor, etc.

tests/ckeditor.jsx
pomek: Ok, I checked that `MainComponent.Children` works.

pomek: However, it looks bad if you look at the tree:

```html
<CKEditor.Context>
    <CKEditor>
```

The names suggest that `Context` is a child of `CKEditor`. In fact, `CKEditor` is a child of `Context`.

pomek: I guess we need to modify the API exported by the package.

```js
import { CKEditor, Context } from '@ckeditor/ckeditor5-react';
```

Looks good. We could also export the `CKEditor` component as default so no breaking changes will be introduced.

```js
// The imports below should works as well:
import CKEditor, { Context } from '@ckeditor/ckeditor5-react';
import CKEditor from '@ckeditor/ckeditor5-react';
```

tests/ckeditor.jsx
pomek: You're right but from the end-user point of view, it does not make any sense to call the handler more than once. We need to figure out how to disable the `onInit()` callback after the first calling.

If we can't use `once()`, we need to use an additional flag.

tests/ckeditor.jsx
ma2ciek: I've added the new package entry point.

tests/ckeditor.jsx
ma2ciek: > You're right but from the end-user point of view, it does not make any sense to call the handler more than once.

I disagree. It depends...

README.md
ma2ciek: Unfortunately, we can't, but maybe it's doable with http-server.

tests/ckeditor.jsx
pomek: > I disagree. It depends...

From what?

ma2ciek: It depends on the usage. Tell me what you'd do on the `onInit` call. Anyway, that won't be a regression as earlier the editor would just crash.

ma2ciek: Maybe I could introduce the `onReady` event for both - editor and context? If the semantics isn't clear with the `onInit`.

tests/context.jsx
ma2ciek: It shouldn't be called. An unhandled promise rejection doesn't make the test failing.

tests/ckeditor.jsx
pomek: > Tell me what you'd do on the onInit call.

We can't assume that it makes sense for somebody. At this moment the handler is called once. After your changes, the handler can be called more than once which is a major breaking change that isn't mentioned in the commit message.

A first example, what I could do in the handler. Let's assume my page has a loading spinner that covers the editor area. I would like to remove the spinner when the editor is ready. After calling the handler for the second time, the script will try to remove the non-existing element and most probably end with an error.

pomek: > Anyway, that won't be a regression as earlier the editor would just crash.

The user knew about the crash. He knew he must refresh the page and was aware that when refreshing, all callbacks will be called once again.

Watchdog "refresh" the editor in the background and calls functions that aren't supposed to be called.

src/ckeditor.jsx
```diff
@@ -51,56 +71,83 @@ export default class CKEditor extends React.Component {
...
+						}
+					} );
+
+					// The onInit should be fired once the `editor` property
+					// can be reached from the <ckeditor> component.
+					// Ideally this part should be moved to the watchdog item creator listeners.
+					setTimeout( () => {
+						if ( this.props.onInit ) {
+							this.props.onInit( this.editor );
+						}
```
pomek, line 111: The `timeout` solution is just bad.
Let's move the callback execution after creating an editor by Watchdog.

```js
this.watchdog.create( this.domContainer.current, this._getConfig() )
  .then( /* the code goes here */ ) 
  .catch( onError );
```

tests/ckeditor.jsx
pomek: However, there is also a counter use case: what about Decoupled Editor and its UI? It must be added each time when the editor is ready.

src/ckeditor.jsx
ma2ciek: We agreed to change `onInit` to `onReady`.

tests/ckeditor.jsx
ma2ciek: We agreed to change `onInit` to `onReady` and call it multiple times.

samples/index.html
```diff
@@ -38,10 +38,10 @@ <h2>Sample</h2>
 			React.createElement( CKEditor, {
 				editor: ClassicEditor,
 				data: editorContent,
-				onInit: editor => {
+				onReady: ( { editor } ) => {
```
pomek, line 41: I have a strong feeling that we should not modify the API of the `onInit`/`onReady` hook.

If the callback is called with an editor instance only, I guess we can use the current (v1) API.

src/ckeditor.jsx
```diff
@@ -5,24 +5,67 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import EditorWatchdog from '@ckeditor/ckeditor5-watchdog/src/editorwatchdog';
+import uid from '@ckeditor/ckeditor5-utils/src/uid';
+import { ContextWatchdogContext } from './context.jsx';
+import ContextWatchdog from '@ckeditor/ckeditor5-watchdog/src/contextwatchdog';
```
pomek, line 11:
```suggestion
import EditorWatchdog from '@ckeditor/ckeditor5-watchdog/src/editorwatchdog';
import ContextWatchdog from '@ckeditor/ckeditor5-watchdog/src/contextwatchdog';
import uid from '@ckeditor/ckeditor5-utils/src/uid';
import { ContextWatchdogContext } from './context.jsx';
```

pomek: I changed the order of imports. External packages (sorted A-Z), then local files.

src/ckeditor.jsx
```diff
@@ -5,24 +5,67 @@
...
+
+		/**
+		 * An instance of EditorWatchdog if the context watchdog was not passed via React's context. `null` otherwise.
+		 */
+		this.watchdog = null;
+
+		/**
+		 * An instance of ContextWatchdog passed from the React's context or `null` otherwise.
+		 */
+		this.contextWatchdog = null;
```
pomek, line 29: Looks like those variables keep the same value. The only difference between them is the source. The first one comes from the CKEditor component itself, the second one comes with CKEditor properties.

Couldn't we have a single (protected) property for a case when the component itself creates the context?

I'd add to a getter:

```js
get watchdog() {
	return this._watchdogContext || this.props.context; // I am unsure about names.
};
```

src/ckeditor.jsx
```diff
@@ -5,24 +5,67 @@
...
+
+		/**
+		 * A unique id for an editor component.
```
pomek, line 32: Comment on why it's needed would be nice.

src/ckeditor.jsx
```diff
@@ -5,24 +5,67 @@
...
+		this._id = uid();
+	}
+
+	get editor() {
+		if ( this.watchdog ) {
+			return this.watchdog.editor;
+		}
+
+		if ( this.contextWatchdog ) {
+			// TODO - try/catch should not be necessary as `getItem` could return `null` instead of throwing errors.
```
pomek, line 43: I guess there is no ticket for this part of the code: https://github.com/ckeditor/ckeditor5/blob/3e6ea99fe28225c52092b621c3593748bb1c168e/packages/ckeditor5-watchdog/src/contextwatchdog.js#L418-L433

If you are going to change it in the future, would be good to attach a ticket id as a reference. Otherwise, I'd remove the TODO.

src/ckeditor.jsx
```diff
@@ -5,24 +5,67 @@
...
+			} catch ( err ) {
+				return null;
+			}
+		}
+
+		return null;
 	}
 
-	// This component should never be updated by React itself.
+	// This component should almost never be updated by React itself.
```
pomek, line 54:
```suggestion
	// The CKEditor component should not be updated by React itself.
	// However, if the component identifier has changed, the whole structure should be created once again.
```

src/ckeditor.jsx
```diff
@@ -5,24 +5,67 @@
...
 		}
 
-		if ( this._shouldUpdateContent( nextProps ) ) {
+		// Only when the editor component changes the whole structure should be restarted.
+		if ( nextProps.id !== this.props.id ) {
+			this._destroyEditor();
+			this._initializeEditor();
+
+			return true;
+		}
```
pomek, line 66: I understand what we want to do here. But I am unsure should the component be updated by React? I guess, manually destroying and creating a new editor should be fine.

Also:
- `Editor#destroy()` returns a promise. Shouldn't we wait for it?
- Same with `Editor#create()`.
- I'd rename `props.id` to something other. Maybe `editorId`. It can be confusing to set the `id` attribute:

```jsx
<CKEditor [...] id="my-ckeditor-element">
```

As for me, I'd expect that the editor container will be marked with the specified identifier.

src/ckeditor.jsx
```diff
@@ -51,60 +94,97 @@ export default class CKEditor extends React.Component {
...
-				}
-
-				const modelDocument = editor.model.document;
-				const viewDocument = editor.editing.view.document;
-
-				modelDocument.on( 'change:data', event => {
-					/* istanbul ignore else */
-					if ( this.props.onChange ) {
-						this.props.onChange( event, editor );
+		const creator = ( el, config ) => {
```
pomek, line 97: What about moving the creator function outside the component as a factory function?

```js
function creatorFactory( element, editorConfig, componentProperties ) {
    return ( el, config ) { /* ... */ };
}
```

Thanks to that, the `_initializeEditor()` body will be a little cleaner.

src/ckeditor.jsx
```diff
@@ -51,60 +94,97 @@ export default class CKEditor extends React.Component {
...
+					viewDocument.on( 'blur', event => {
+						/* istanbul ignore else */
+						if ( this.props.onBlur ) {
+							this.props.onBlur( event, editor );
+						}
+					} );
+
+					// The `onReady` callback should be fired once the `editor` property
+					// can be reached from the `<ckeditor>` component.
+					// Ideally this part should be moved to the watchdog item creator listeners.
```
pomek, line 130: Is it possible now? If so, let's do it.

src/ckeditor.jsx
```diff
@@ -51,60 +94,97 @@ export default class CKEditor extends React.Component {
...
 
-				onErrorCallback( error );
+			onErrorCallback( {
+				error,
+				phase,
+				willEditorRestart: causesRestart
 			} );
+		};
+
+		// TODO: Is it better to use instanceof or duck typing?
```
pomek, line 151: The end bundle will contain both files anyway so we can use the `instanceof` check.

```suggestion
```

src/ckeditor.jsx
```diff
@@ -51,60 +94,97 @@ export default class CKEditor extends React.Component {
...
-				const onErrorCallback = this.props.onError || console.error;
+		const onError = ( { error, phase = 'runtime', causesRestart = false } ) => {
+			const onErrorCallback = this.props.onError || console.error;
 
-				onErrorCallback( error );
+			onErrorCallback( {
+				error,
+				phase,
+				willEditorRestart: causesRestart
 			} );
```
pomek, line 148:
```suggestion
			onErrorCallback( error, {
				phase,
				willEditorRestart: causesRestart
			} );
```

WDYT? If we won't modify the API of `onReady` event, users after upgrading the package will have less to fix. But it's just thought, we can leave it.

src/context.jsx
```diff
@@ -0,0 +1,101 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import ContextWatchdog from '@ckeditor/ckeditor5-watchdog/src/contextwatchdog';
+
+export const ContextWatchdogContext = React.createContext( 'contextWatchdog' );
```
pomek, line 10: Is it safe if a user will use two `<Context>` components on the same page?

pomek: If so, WDYT about adding a getter to the Context component that will return the value?

src/ckeditor.jsx
```diff
@@ -137,13 +217,15 @@ export default class CKEditor extends React.Component {
...
 
+CKEditor.contextType = ContextWatchdogContext;
+
 // Properties definition.
 CKEditor.propTypes = {
 	editor: PropTypes.func.isRequired,
 	data: PropTypes.string,
 	config: PropTypes.object,
 	onChange: PropTypes.func,
-	onInit: PropTypes.func,
```
pomek, line 146:
```js
  onInit: function(props, propName, componentName) {
      return new Error( 'The "onInit" callback is not supported anymore. Use the "onReady" property instead.' );
  },
```

WDYT?

src/index.js
```diff
@@ -0,0 +1,6 @@
+import CKEditor from './ckeditor.jsx';
```
pomek, line 1: We need a test for the file since we're changing the output (`dist/ckeditor.js` file). It will not export the default component anymore.

The test should check whether `dist/index.js` exports an object and whether the object has two properties with proper types.

pomek: Also, the file misses the license banner.

src/index.js
```diff
@@ -0,0 +1,6 @@
...
+import Context from './context.jsx';
+
+export default CKEditor;
```
pomek, line 4:
```suggestion
```

As we talked via priv, it won't work so let's drop that.

pomek (changes requested):
We need to decide about the AIP in the `onReady`/`onError` callbacks.

Also, minor refactoring would be nice.

Missing tests (and dropped coverage) could be fixed in a separate issue.

src/ckeditor.jsx
ma2ciek: I was thinking about a similar way. The proposed solution won't work as both solutions have slightly different API in case of getting the editor (as one watchdog is for the editor and the second is for the context.

README.md
ma2ciek: Outdated.

src/ckeditor.jsx
ma2ciek: I've just written a small adapter for the Context Watchdog feature. I hope you'll like it.

src/context.jsx
ma2ciek: This context uses later a provider. So in theory you could use such structure:

```
<Context>
  <Editor />
</Context>
<Context>
  <Editor />
</Context>
```

And everything should work.

src/ckeditor.jsx
ma2ciek: :+1:

README.md
```diff
@@ -57,6 +57,12 @@ Build a minified version of the package that is ready to publish:
...
+
```
pomek, line 65:
```suggestion
```

src/ckeditor.jsx
```diff
@@ -51,19 +80,29 @@ export default class CKEditor extends React.Component {
 	}
 
 	_initializeEditor() {
-		this.props.editor
-			.create( this.domContainer.current, this._getConfig() )
-			.then( editor => {
-				this.editor = editor;
+		if ( this.context instanceof ContextWatchdog ) {
+			this.watchdog = new ContextWatchdogToEditorWatchdogAdapter( this.context );
```
pomek, line 84: We need to find a better name for that adapter. Couldn't it be `EditorWatchdogAdapter`?

src/ckeditor.jsx
```diff
@@ -87,24 +126,26 @@ export default class CKEditor extends React.Component {
...
 
-				onErrorCallback( error );
+				// The `onReady` callback should be fired once the `editor` property
+				// can be reached from the `<CKEditor>` component.
+				// Ideally this part should be moved to the watchdog item creator listeners.
+				setTimeout( () => {
+					if ( this.props.onReady ) {
+						this.props.onReady( this.editor );
+					}
+				} );
```
pomek, line 137: At this stage, `this.editor` should always return something since `this.watchdog` exists. Could we get rid of the timeout?

src/ckeditor.jsx
```diff
@@ -137,21 +178,80 @@ export default class CKEditor extends React.Component {
 	}
 }
 
+/**
+ * An adapter aligning Context Watchdog API to the Editor Watchdog API for easier usage.
+ */
+class ContextWatchdogToEditorWatchdogAdapter {
```
pomek, line 184: Some JSDoc comments such as `@params` and `@returns` would be nice when somebody in the future will have to edit this file.

src/ckeditor.jsx
pomek: That's good.

src/context.jsx
pomek: 👍

src/ckeditor.jsx
ma2ciek: Actually it clearly says what it keeps inside and what interface it matches.

pomek: `AtSomePointWeWillWriteAStoryAsAClassName` because it's _clearly says what it keeps inside and what interface it matches._

I'd suggest to simplify it and add proper docs to the class.

src/ckeditor.jsx
ma2ciek: Added.

src/ckeditor.jsx
ma2ciek: It's not possible for now.

pomek (approved):
LGTM.

PR #213, opened 2021-01-20
Fix: Add React 17 as a valid peerdependency.

fedemp:
Listing only React 16 creates a problem with React 17.

`npm ls react` reports that both React versions are missing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "webpack-cli": "^3.3.12"
   },
   "peerDependencies": {
-    "react": "^16.13.1"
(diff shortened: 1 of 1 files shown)


fedemp:
Screenshot from 2021-01-20 10-57-39

pomek:
Thanks for pointing this out. We went this way https://github.com/ckeditor/ckeditor5-react/blob/cc8855661a81433eb5bb4035a5a4dd51df7efb45/package.json#L61

The change is available as `3.0.1` – https://github.com/ckeditor/ckeditor5-react/releases/tag/v3.0.1

PR #272, opened 2021-11-26
Updated required node version to 14

przemyslaw-zan:
Other: Updated the required version of Node.js to 14. See ckeditor/ckeditor5#10972.

MAJOR BREAKING CHANGE: Upgraded the minimal versions of Node.js to `14.0.0` due to the end of LTS.


pomek (approved):
LGTM.

PR #284, opened 2022-02-08
Use async/await in CKEditorContext#_destroyContext() to fix the destruction process

### LINKED ISSUE #283: CKEditorContext component doesn't destroy properly
## Description

It seems that `<CKEditorContext>` component doesn't destroy properly, as it keeps an open WebSocket connection after unmounting. For example, as a result, it leaves connected users in the presence list. 

The easiest way to reproduce the problem is to use our [Comments outside of the editor for React sample](https://github.com/ckeditor/ckeditor5-collaboration-samples/tree/master/real-time-collaboration-comments-outside-of-editor-for-react), where Context + RTC are integrated. The sample should be modified to allow component destruction. I have prepared a simple solution that adds a button for toggling editors - it could be used for the reproduction, just replace the original `sample.js` file with the one from [my gist](https://gist.github.com/Mgsy/c18ba7624983bf8ab52f93de967d37b4).

After creating the editors, the WebSocket connection is open, but after removing the editor, it is still open. **This issue doesn't occur with the default `<CKEditor>` component or default Context integration.**

## Possible solution

Here is the part responsible for destroying the context after unmounting the component:

[https://github.com/ckeditor/ckeditor5-react/blob/master/src/ckeditorcontext.jsx#L82-L87](https://github.com/ckeditor/ckeditor5-react/blob/master/src/ckeditorcontext.jsx#L82-L87)

```javascript
_destroyContext() {
	if ( this.contextWatchdog ) {
		this.contextWatchdog.destroy();
		this.contextWatchdog = null;
	}
}
```

I noticed that converting this method to the async method fixes the issue and makes all asynchronous code from ContextWatchdog run properly (however, I don't know if it is a correct approach).

```javascript
async _destroyContext() {
	if ( this.contextWatchdog ) {
		await this.contextWatchdog.destroy();
		this.contextWatchdog = null;
	}
}
```

**Before the change** (see the dev tools and _Time_ column of the WebSocket connection - it's _pending_ after the destroy):

**After the change**:

Mgsy:

Fix: Use async/await in CKEditorContext#_destroyContext() to handle context destruction properly. Closes #283.

---

It is not that easy to test, so here is a short guide explaining how to do it:

1. Get the `real-time-collaboration-comments-outside-of-editor-for-react` collaboration sample and replace original `src/sample.js` with [my code](https://gist.github.com/Mgsy/9453453fd8b95ab97bf5ff8ab94d2857).
2. In `src/sample.js` add import path to the local `ckeditor5-react` package.
3. Build and run the sample.
4. Open the dev tools and the network tab.
5. Click the button to create editors.
6. The WebSocket connection should be _pending_.
7. Click the button to destroy editors.
8. The WebSocket connection should be closed.

The better description and GIFs are available in the original issue.

### DIFF
--- a/src/ckeditorcontext.jsx
+++ b/src/ckeditorcontext.jsx
@@ -79,9 +79,9 @@ export default class CKEditorContext extends React.Component {
 		} );
 	}
 
-	_destroyContext() {
+	async _destroyContext() {
 		if ( this.contextWatchdog ) {
-			this.contextWatchdog.destroy();
+			await this.contextWatchdog.destroy();
 			this.contextWatchdog = null;
 		}
 	}
--- a/tests/ckeditorcontext.jsx
+++ b/tests/ckeditorcontext.jsx
@@ -119,6 +119,36 @@ describe( '<CKEditorContext> Component', () => {
 			expect( editorCreateSpy.firstCall.args[ 1 ].initialData ).to.equal( '<p>Foo</p>' );
 			expect( editorCreateSpy.secondCall.args[ 1 ].initialData ).to.equal( '<p>Bar</p>' );
 		} );
+
+		it( 'should wait for the `ContextWatchdog#destroy()` promise when destroying the context feature', async () => {
+			let watchdog;
+
+			await new Promise( ( res, rej ) => {
+				wrapper = mount(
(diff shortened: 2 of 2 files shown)


pomek (commented):
Post-merge LGTM.

PR #300, opened 2022-04-05
Support for the lock mechanism

pomek:

Other: Aligned the `<CKEditor>` component API to use the new lock mechanism when enabling/disabling the read-only mode.

MAJOR BREAKING CHANGE: Due to introducing the lock mechanism for the `Editor#isReadOnly` property, the `<CKEditor>` component uses the new way of enabling the read-only mode in the editor. The component requires an instance of CKEditor 5 in version 34 or higher. See ckeditor/ckeditor5#10496.

---

### Additional information

* The sample is adjusted to use the new API. Hence, it requires bumping CKEditor 5 dependencies to v34.0.0.


pomek:
## "Testing phase"

Let's start by checking the current version (when I write this comment, it is CKEditor 5 v33), then simulate a new CKEditor 5 release.

### Steps for testing v33.0.0

* Reinstall modules to use the latest dependencies.
* Build the component (`yarn run build`).
* Temporary revert this change: https://github.com/ckeditor/ckeditor5-react/pull/300/files#diff-04f91672b4fd43492d3d43884abab0536853d5052d74a28b0b807f57a4082983R87 as it requires CKEditor 5 v34.
* From the root directory, call the HTTP server.
* Open the `sample/` directory via HTTP.
* Switch the editor to read-only mode. It should work.
* You can add the following line between these two: https://github.com/ckeditor/ckeditor5-react/blob/8afd898c59a982ea36029355f688cf1f44c7e489/sample/index.html#L51-L52
  ```
  disabled: true
  ```
* Refresh the page. The editor should be disabled by default.

### Steps for testing v34+

* From the CKEditor 5 directory, checkout to the `#master` branch and pull changes.
* Update [the hardcoded version](https://github.com/ckeditor/ckeditor5/blob/ffebb42a79055d5563f901a83ee7f2946324b4a7/packages/ckeditor5-utils/src/version.js#L14) from `33` to `34`.
* Go to `packages/ckeditor5-build-classic` and call `yarn run build`
* In the `ckeditor5-react` project, find the `@ckeditor/ckeditor5-build-classic` package in the `node_modules/` directory and replace its content with the fresh build.
* Revert the "_Temporary revert this change_" point.
* Refresh the sample. If you have the `disabled:true` option in `index.html`, you cannot disable the read-only mode using buttons. You need to clear the integration lock.
  ```js
  editor.disableReadOnlyMode( 'Lock from React integration (@ckeditor/ckeditor5-react)' );
  ```

pomek:
We did a live review and it looks good. Before merging the PR, we need to bump CKEditor 5 to 34.0.0.

PR #342, opened 2022-11-10
Enforce cleanup completion before subsequent editor initialization

### LINKED ISSUE #338: Editor not displaying at first load with Strictmode
Hello team !

Thank you for your hard work :)

I noticed that my editor wasn't displayed when I start my server (with npm start). To make it showing, I need to reload the page on my navigator, and then it's working fine.

After multiple hours of research I found out that it was occurring because of the React StrictMode. This also comes because I retrieve some data from my own server, and I use _useEffect_ for waiting that all data is loaded before displaying anything.

You can reproduce it in a new React App, with just this code:

```jsx
// App.js:
import React, { useEffect, useState } from "react";
import ClassicEditor from "@ckeditor/ckeditor5-build-classic";
import { CKEditor } from "@ckeditor/ckeditor5-react";

function App() {
  const [isLoading, setIsLoading] = useState(true);
  var someData = "notEmpty";

  useEffect(() => {
    setTimeout(function () {
      someData && setIsLoading(false);
    }, 1000); // Simulate some data loading from a server
  }, [someData]);

  return (
    <>
      {isLoading ? (
        <i>Logo loading</i>
      ) : (
        <>
          <div className="App">
            <CKEditor
              editor={ClassicEditor}
              data="<p>Hello from CKEditor 5!</p>"
            />
          </div>
        </>
      )}
    </>
  );
}

export default App;
```

```jsx
// index.js:
`import React from 'react';
import ReactDOM from 'react-dom/client';
import './index.css';
import App from './App';
import reportWebVitals from './reportWebVitals';

const root = ReactDOM.createRoot(document.getElementById('root'));
root.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);

// If you want to start measuring performance in your app, pass a function
// to log results (for example: reportWebVitals(console.log))
// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
reportWebVitals();
```

Hope you can reproduce it. If you have any questions, do not hesitate;

Thank you for all !

corymharper:
Fix: Fixed editor initialization running before a subsequently initiated destroy was completed. Closes #338.

---

### Additional information

These problems may have only become possible after the introduction of asynchronous behavior in #310 . In testing, it was evident that sometimes initializations of the editor were happening before cleanup from the unmounting phase completed, this is made possible by an asynchronous context. When that would occur, the initialization would detect an already present watchdog instance, and abort creating anew watchdog and paired editor instance. Subsequently the interrupted cleanup would continue and destroy the existing editor, leaving an empty div where the editor should be on the DOM. I haven't been able to replicate this issue using the `sample/index-18.html` file, so its probably only evident in larger applications.

A roadblock I encountered after running the tests was that there seemed to be occasions where the watchdog's destruction didn't return a promise, I wasn't expecting that, but I remedied it by checking it was a promise before calling `.then` on it.

I was able to remove a check for the watchdog existing that was added in #310, because waiting for pending destructions means that check is always passed.

### DIFF
--- a/src/ckeditor.jsx
+++ b/src/ckeditor.jsx
@@ -19,12 +19,12 @@ export default class CKEditor extends React.Component {
 		super( props );
 
 		/**
-		 * While cleanup from unmounting is in progress, contains a promise that resolves when cleanup is complete.
+		 * While cleanup from destroying an editor is in progress, contains a promise that resolves when cleanup is complete.
 		 * Otherwise it contains null.
 		 *
 		 * @type {Promise|null}
 		 */
-		this.cleanupInProgress = null;
+		this.editorDestructionInProgress = null;
 
 		// After mounting the editor, the variable will contain a reference to the created editor.
 		// @see: https://ckeditor.com/docs/ckeditor5/latest/api/module_core_editor_editor-Editor.html
@@ -123,13 +123,7 @@ export default class CKEditor extends React.Component {
 	 * @returns {Promise}
  	 */
 	async componentWillUnmount() {
-		this.cleanupInProgress = new Promise( resolve => {
-			this._destroyEditor().then( () => {
-				resolve();
-			} );
-		} ).then( () => {
-			this.cleanupInProgress = null;
-		} );
+		await this._destroyEditor();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


pomek:
Incredible work! We will look at the PR next week. Hopefully, I'll be able to write an integration test that prevents introducing a regression in the future.

PR #348, opened 2022-11-29
Do not attach husky hooks when installing a repository as a dependency (instead of cloning)

martnpaneq:
Internal: Do not attach husky hooks when installing a repository as a dependency (instead of cloning)


pomek (approved):
LGTM.

PR #367, opened 2023-03-23
Migration to TS

niegowski:

Feature: Migrate package to TypeScript. Closes ckeditor/ckeditor5#13542.

MAJOR BREAKING CHANGE: Integration now requires Editor version 37 or later.


src/ckeditor.tsx
```diff
@@ -99,50 +98,40 @@ export default class CKEditor extends React.Component {
...
 	/**
 	 * Re-render the entire component once again. The old editor will be destroyed and the new one will be created.
-	 *
-	 * @returns {Promise}
 	 */
-	async componentDidUpdate() {
-		await this._destroyEditor();
-		await this._initializeEditor();
+	public override componentDidUpdate(): void {
+		Promise.resolve()
```
martnpaneq, line 110: *minor* Turns out that `async` syntax is working fine. It looks a bit cleaner
```
public override async componentDidUpdate(): Promise<void> {
	await this._destroyEditor();
	await this._initializeEditor();
}
```

src/ckeditor.tsx
```diff
@@ -99,50 +98,40 @@ export default class CKEditor extends React.Component {
...
 	}
 
 	/**
 	 * Destroy the editor before unmounting the component.
-	 *
-	 * @returns {Promise}
  	 */
-	async componentWillUnmount() {
-		await this._destroyEditor();
+	public override componentWillUnmount(): void {
```
martnpaneq, line 118: removing `async` is risky here too

src/ckeditor.tsx
```diff
@@ -295,45 +274,104 @@ export default class CKEditor extends React.Component {
...
+			if ( props[ propName ] ) {
+				return new Error(
+					'The "onInit" property is not supported anymore by the CKEditor component. Use the "onReady" property instead.'
+				);
+			}
+		}
+	};
+
+	// Default values for non-required properties.
+	public static defaultProps: Partial<Props<Editor>> = {
```
martnpaneq, line 304: Unfortunately defaultProps disable the intellisense for required properties.

For example, with defaultProps, IDE is not showing that editor property is missing:
image

Without defaultProps:
image

I tried fixing it, but tests fail and the fix is not easy.

martnpaneq (approved):

pomek (commented):
As I am not in the process, I do not understand why we have two samples:

1. `sample/`
2. `demo-*/` 

I can guess that the first is just a sample using the component and some CKEditor 5 build, while the second is a demo application (more realistic scenario). It would be good to describe both in README and explain which should be checked before releasing a new version. Maybe we do not need the `sample/` directory anymore. Instead, we should develop example applications that could be checked using E2E tests on CI.

pomek:
WDYT about dropping support for non-supported React versions? It means the CKEditor component will require React 18.0+.

image

https://endoflife.date/react

pomek (approved):
LGTM.

Issue #351, opened 2022-12-15
CKEditor5 Jest Testing - TypeError: Cannot read properties of null (reading 'model')


jiefugong:
Hi,

I'm using the CKEditor5 React Component with one of the ready-to-use editor builds -- specifically, the `ClassicEditor`. I've installed both of these using `yarn` in my local project and have code that is 100% identical to the example given on [this quick start page](https://ckeditor.com/docs/ckeditor5/latest/installation/frameworks/react.html#the-javascript-heap-out-of-memory-error)

So far, using the editor itself has been fine in my application. However, when I attempt to write Jest tests for this component I get the following error:

```
TypeError: Cannot read properties of null (reading 'model')
    at model (@ckeditor/ckeditor5-react/dist/webpack:/CKEditor/node_modules/@ckeditor/ckeditor5-watchdog/src/editorwatchdog.js:241:12)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

My Jest test is incredibly simple:

```
describe('TextEditor', () => {
  it('renders successfully', () => {
    expect(() => {
      render(<TextEditor />)
    }).not.toThrow()
  })
})
```

I've tried to find similar questions online but cannot find one for this particular TypeError. I also am assuming that since this is a ready-to-use editor build that it should work out of the box -- I'm not sure if some sort of Webpack configuration is required for this. Is there anything I'm doing wrong or is this perhaps a known issue?

Aerophite:
I also have this issue. Does anyone have any insight into how to fix it?

Aerophite:
@jiefugong 

This appears to be fixed in version 5.0.6

pomek:
Sounds good. @jiefugong, could you confirm?

alexcoliveira1:
I currently have that same error, running on v5.0.6, started right after I upgraded from v5.0.2.

2-fly-4-ai:
Have this same issue working in next.js. I'm not running any tests, just using the editor like normal. Everything works fine, saves to the db but I can't stop this error from popping afterwards. 

"The config.initialData option cannot be used together with initial data passed as the first parameter of [Editor.create()](https://ckeditor.com/docs/ckeditor5/latest/api/module_core_editor_editor-Editor.html#static-function-create)."

kyrofa:
[...]

PR #383, opened 2023-06-02
CKEditor: handle being torn down before initialization

kyrofa:
It's easy, particularly in tests, to end up rendering the editor only to then quickly change a prop that causes it to no longer render. It's possible for this to happen quickly enough that `_initializeEditor()` hasn't finished running yet, thus ending up in `_createEditor()` with a null `element`. That ends up causing CKEditor itself to throw an exception, and ends up not defining the editor, which then makes the watchdog throw an exception. The latter error is not catchable in an error boundary due to its asynchronous nature.

Fix this by making `_createEditor()` a little more defensive: make sure we have an element before bothering to create an editor. If we don't have one, we know we're in the midst of a teardown, so create a dummy editor so the watchdog has something to destroy.

This might be related to #351, but I'm not certain.

### DIFF
--- a/src/ckeditor.tsx
+++ b/src/ckeditor.tsx
@@ -166,6 +166,15 @@ export default class CKEditor<TEditor extends Editor> extends React.Component<Pr
 	 * @param config CKEditor 5 editor configuration.
 	 */
 	private _createEditor( element: HTMLElement | string | Record<string, string>, config: EditorConfig ): Promise<TEditor> {
+		if ( !element ) {
+			// In our case, the only way this element is falsy is if we're caught in
+			// the middle of tearing down this component. This isn't really an error,
+			// so it doesn't make sense to reject the promise-- that would make the
+			// watchdog explode anyway. Just create an empty editor so the watchdog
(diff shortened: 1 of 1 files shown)


Witoso:
@pomek will your team take a look?

kyrofa:
FYI, I did sign the CLA, not sure why your system things I haven't.

Screenshot from 2023-08-17 10-14-56

Witoso:
@vokiel could you check if it's signed?

vokiel:
Check, all good :heavy\_check\_mark:

kyrofa:
Any chance someone could take a look at this? @pomek, friendly ping.

Mati365:
@Witoso / @kyrofa This PR can be closed. We fixed this issue [here](https://github.com/ckeditor/ckeditor5-react/commit/ce6cad06e7a847b48189ddfaeed57921ce314874).

pomek:
Closing due to https://github.com/ckeditor/ckeditor5-react/pull/383#issuecomment-2210212439.

Issue #373, opened 2023-04-28
Allow disabling watchdog in the component


Mgsy:
## Description

This feature request is about introducing a possibility for removing Watchdog integration from the component.

pomek:
### Scope

- Add support for a new property called `disableWatchdog`
- By default it should be set to `false`
- It set to `true`, let's call [`_createEditor()`](https://github.com/ckeditor/ckeditor5-react/blob/9e123bc4f626a66c53d77fcc4b73af4f549cb4f9/src/ckeditor.tsx#L168) method and skip [the entire watchdog initialization phase](https://github.com/ckeditor/ckeditor5-react/blob/9e123bc4f626a66c53d77fcc4b73af4f549cb4f9/src/ckeditor.tsx#L136-L159).

glynam1:
@pomek / @Mgsy 

Is this expected to work within a CKEditor Context? 
It appears that disabling watchdog within an editor that is contained within a CKEditor Context throws a soft plugin required error because the CloudServices plugin is not available  . 

It almost appears that the CloudServices plugin is implicitly provided through the watchdog coming from the context

jake-netflix:
> It appears that disabling watchdog within an editor that is contained within a CKEditor Context throws a soft plugin required error because the CloudServices plugin is not available .

I ran into the same `CloudServices` soft required. Can't use `disableWatchdog` yet because my editors are in a context.

PR #388, opened 2023-06-07
Internal: Aligned release tools to the latest CKEditor 5 dev.

martnpaneq:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Internal: Aligned release tools to the latest CKEditor 5 dev. Closes ckeditor/ckeditor5-react#378.

---

### Additional information

*For example – encountered issues, assumptions you had to make, other affected tickets, etc.*

PR #392, opened 2023-06-19
Added `disableWatchdog` property to the React component.

martnpaneq:

Feature: Added `disableWatchdog` property to the React component. Closes ckeditor/ckeditor5-react#373.

---

### Additional information


przemyslaw-zan (approved):
`yarn.lock` seems to have leaked from #388, so it should not be commited here. Other than that, LGTM

martnpaneq:
> `yarn.lock` seems to have leaked from #388, so it should not be commited here. Other than that, LGTM

Good catch, fixed.

pomek (approved):
LGTM.

PR #412, opened 2023-09-13
Node version bump

przemyslaw-zan:
Other: Updated the required version of Node.js to 18 when developing the repository. See ckeditor/ckeditor5#14924.


pomek (approved):
LGTM.

PR #461, opened 2024-03-22
Allow creation for context without a contextwatchdog

### LINKED ISSUE #409: Option to disable watchdog on <CKEditorContext>
Similar to #373, it would be great if we could disable watchdog on the `<CKEditorContext>` component, so that there's no context watchdog running.

As an example, watchdog was scanning 100k unique objects (over 400k references to those objects) and then generating errors, because some of those objects are a `Proxy` that handle `get` and throw if a key is unexpected. Since this happens during `_create`, it fails to initialize altogether.

(It does this when it calls `getSubNodes`)
https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-watchdog/src/utils/getsubnodes.ts

As a workaround, I use this:
```tsx
<CKEditorContext context={ContextWithoutWatchdog} ... > ... </CKEditorContext>
```

And this class:
```ts
export class ContextWithoutWatchdog extends Context {
  public _watchdogExcluded = true;
}
```

glynam1:
Feature (ckeditor5-react): Allowing creation of a ckeditor5 context without a context watcdhog Closes #409.

### DIFF
--- a/src/ckeditorcontext.tsx
+++ b/src/ckeditorcontext.tsx
@@ -67,6 +67,9 @@ export default class CKEditorContext<TContext extends Context = Context> extends
 	}
 
 	private async _initializeContextWatchdog( config?: ContextConfig ): Promise<void> {
+		if ( this.props.disableWatchdog ) {
+			return;
+		}
 		this.contextWatchdog = new ContextWatchdog( this.props.context!, this.props.watchdogConfig );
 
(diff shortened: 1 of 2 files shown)


glynam1:
@Witoso pinging you on this PR because I see you've been active on the ticket that it references. I don't believe this CI failure is due to this PR. 
It looks like nightly builds have been failing for a while?
https://app.circleci.com/pipelines/github/ckeditor/ckeditor5-react

PR #470, opened 2024-04-18
Fixed resolving the peer dependencies

DawidKossowski:

Internal: Fixed resolving the peer dependencies.

---

### Additional information

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -11,20 +11,20 @@ const path = require( 'path' );
 const webpack = require( 'webpack' );
 const { bundler } = require( '@ckeditor/ckeditor5-dev-utils' );
 const TerserPlugin = require( 'terser-webpack-plugin' );
+const { dependencies, peerDependencies } = require( './package.json' );
+
+const externals = Object.keys( { ...dependencies, ...peerDependencies } ).reduce( ( acc, currentValue ) => {
+	acc[ currentValue ] = currentValue;
+
+	return acc;
+}, {} );
 
 module.exports = {
 	context: __dirname,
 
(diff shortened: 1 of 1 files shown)


pomek:
@DawidKossowski, could you give a short explanation of why this change is needed?

filipsobol (approved):

DawidKossowski:
Sure. Now, every package used in the source code is included in the build, even if it's listed in `peerDependencies`. For instance, we use Watchdog version `40.1.0` in React integration, which is also listed in the peer dependencies. However, it gets bundled into the file: https://cdn.jsdelivr.net/npm/@ckeditor/ckeditor5-react/dist/index.js, since it is not marked as external in the Webpack config. This leads to the issue that if the integrator installs Watchdog version `41.3.1`, it won't matter and won't be used since `ckeditor5-react` already includes this package.

Additionally, those packages will be duplicated in `node_modules` with different versions.

pomek:
Thanks. I think the issue touches the Angular integration as well.

pomek (approved):

Reinmar:
> Internal: Fixed resolving the peer dependencies.

Is it really an internal change? Why did we need a change? It's a pity we don't have an actually issue explaining why we're changing this.

[@DawidKossowski's comment](https://github.com/ckeditor/ckeditor5-react/pull/470#issuecomment-2063461771) mention issues with the published bundle, AFAIU. In this case, this PR is a bug fix. And should be mentioned in the changelog.

pomek:
@Reinmar, I will handle it when the package is ready to release.

Issue #442, opened 2024-01-18
TypeError: Cannot read properties of null (reading 'model')


qweyrt:
I'm trying to implement CKEditor 5 to my Nextjs project using this guide: https://ckeditor.com/docs/ckeditor5/latest/installation/integrations/next-js.html

However when I try to run the webpage shows this error: 
![image](https://github.com/ckeditor/ckeditor5-react/assets/81255174/473d07a5-a057-44a6-824b-583b7ea48774)

My code is exactly like in the guide:
custom-editor.tsx:
```
import React from 'react';
import { CKEditor } from "@ckeditor/ckeditor5-react";
import Editor from "ckeditor5-custom-build";

const editorConfiguration = {
    toolbar: [
        'heading',
        '|',
        'bold',
        'italic',
        'link',
        'bulletedList',
        'numberedList',
        '|',
        'outdent',
        'indent',
        '|',
        'imageUpload',
        'blockQuote',
        'insertTable',
        'mediaEmbed',
        'undo',
        'redo'
    ]
};

function CustomEditor( props ) {
        return (
            <CKEditor
                editor={ Editor }
                config={ editorConfiguration }
                data={ props.initialData }
                onChange={ (event, editor ) => {
                    const data = editor.getData();
                    console.log( { event, editor, data } );
                } }
            />
        )
}

export default CustomEditor;
```
page.tsx:
```
'use client' 

import React from 'react';
import dynamic from 'next/dynamic';

const CustomEditor = dynamic( () => {
  return import( '../components/frontend/custom-editor' );
}, { ssr: false } );

function Reports() {
  return (
    <CustomEditor
      initialData='<h1>Hello from CKEditor in Next.js!</h1>'
    />
  );
}

export default Reports;

```

So what happened here? And how do I fix it?

Witoso:
Could you share how `ckeditor5-custom-build` looks in your `package.json`? 

@gorzelinski would you mind verifying if next didn't change something on their side?

nicu-chiciuc:
I'm having the same error. Although I assume the circumstances are slightly different. It'd be hard to provide a full reproduction, but here's my journey.

We have a pnpm monorepo.

When testing an integration with Next.js and CkEditor, I've used the online builder and put the output in a separate package.
This allowed to circumvent the issue of Next.js not allowing Global CSS, which is used in CkEditor plugins.

This worked, and I managed to test some custom plugins.

Then, I've tried to move the code directly in Next.js
[...]

Issue #471, opened 2024-04-18
Quick rerendering causes editor issues


DawidKossowski:
In `StrictMode` or in situations where the application quickly rerenders, the next editor instance is rerendered or attempts to rerender before the previous instance is destroyed, especially when the editor uses async calls (for example, for getting the cloud services token).

We've implemented a mechanism to keep track of the destroying promise (`editorDestructionInProgress`) and to wait for it to resolve before initializing the next editor instance. However, in some cases, the next editor instance is initialized before the previous one starts the destroying process (`editorDestructionInProgress` is null). This results in two editor instances on the page or in an error: `CKEditorError: editor-source-element-already-used`. The outcome depends on the integration.

This issue is non-deterministic and depends on the React versions, but it can be reproduced by adding the config with `cloudServices` property (in our demo: `demo-react-18`) to trigger a request for a token during editor initialization.

DawidKossowski:
@Mati365 please check this mechanism in both integrations: `CKEditor` and `useMultiRootEditor`.

Mati365:
Fixed in https://github.com/ckeditor/ckeditor5-react/releases/tag/v7.0.0. Closing as done.

Issue #469, opened 2024-04-17
onReady is called after <CKEditor> is unmounted


stonebk:
A bit of an edge case I ran into while testing. I have a component that looks something like this:

```tsx
    const [editor, setEditor] = useState<HeadlessEditor>(); 
    const onReadyCallback = useCallback(
      (editor) => {
        setEditor(editor);
      },
      []
    );

    return <CKEditor config={config} editor={MyEditor} onReady={onReadyCallback}  />;
```

I have some small tests that run faster than it takes to initialize the editor, so `<CkEditor>` unmounts before onReady is called, but onReady is called anyways resulting in the following warning:

```
  console.error
    Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

My guess is that this is because [the onReady call happens in a setTimeout](https://github.com/ckeditor/ckeditor5-react/blob/e4db82dad4cf20551eac0357839053d29154020b/src/ckeditor.tsx#L225-L229):

```tsx
				// The `onReady` callback should be fired once the `editor` property
				// can be reached from the `<CKEditor>` component.
				// Ideally this part should be moved to the watchdog item creator listeners.
				setTimeout( () => {
					if ( this.props.onReady ) {
						this.props.onReady( editor );
					}
				} );
```

Would be nice to not have to work around this!

Mati365:
@stonebk Can you check if this issue is still reproducible? We managed to [release](https://github.com/ckeditor/ckeditor5-react/releases/tag/v7.0.0) few modifications in initialization mechanism and upgrading might solve the issue.

stonebk:
> @stonebk Can you check if this issue is still reproducible? We managed to [release](https://github.com/ckeditor/ckeditor5-react/releases/tag/v7.0.0) few modifications in initialization mechanism and upgrading might solve the issue.

I will take a look, thanks!

stonebk:
@Mati365 yay! v7.0.0 looks to have fixed this!

PR #473, opened 2024-04-22
Quick rerendering causes editor issues

Mati365:

Fix: Add CKEditor component initialize / destroy semaphore to prevent race conditions during fast rerender. Closes https://github.com/ckeditor/ckeditor5-react/issues/442, https://github.com/ckeditor/ckeditor5-react/issues/469, https://github.com/ckeditor/ckeditor5-react/issues/471, https://github.com/ckeditor/ckeditor5-react/issues/476.

---

### Additional information

Probably related to these issues: 
https://github.com/ckeditor/ckeditor5-react/issues/469
https://github.com/ckeditor/ckeditor5-react/issues/409
https://github.com/ckeditor/ckeditor5/issues/15980
https://github.com/ckeditor/ckeditor5-react/issues/442
https://github.com/ckeditor/ckeditor5-react/issues/476

### DIFF
--- a/demo-multiroot-react-18/src/App.tsx
+++ b/demo-multiroot-react-18/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { StrictMode, useState } from 'react';
 import MultiRootEditorDemo from './MultiRootEditorDemo';
 import MultiRootEditorRichDemo from './MultiRootEditorRichDemo';
 import ContextMultiRootEditorDemo from './ContextMultiRootEditorDemo';
@@ -43,7 +43,7 @@ export default function App(): JSX.Element {
 	};
 
 	return (
-		<>
+		<StrictMode>
 			<h1>CKEditor 5 – useMultiRootEditor – development sample</h1>
 
 			<div className="buttons" style={ { textAlign: 'center' } }>
@@ -69,6 +69,6 @@ export default function App(): JSX.Element {
 				</button>
 			</div>
 			{ renderDemo() }
-		</>
+		</StrictMode>
 	);
 }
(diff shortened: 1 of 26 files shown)


glynam1:
Screenshot 2024-04-22 at 23 44 18


Maybe I'm misunderstanding the aim of this PR ( or its implementation ) 

But will rerendering editors during instantiation still cause CkeditorContext editors to crash with the above error?

Mati365:
@glynam1 Hmm, I haven’t come across this one before, and it appears that it hasn’t been tested in our coverage. Has there been a reported issue that details the crash?

glynam1:
I had left a comment on this ticket: https://github.com/ckeditor/ckeditor5/issues/15980 as I had hoped that it would have fixed things. It appears that the Context plugin is often left out of these tests (you can't disable it either as per https://github.com/ckeditor/ckeditor5-react/issues/409)

Mati365:
@glynam1 Thank you, I'll check that.

DawidKossowski:
@Mati365, I've noticed that we have a similar issue with the `useMultiRootEditor` hook. When you use `demo-multiroot-react-18`, enabling StrictMode and passing `disableWatchdog: true` into the props, and initializing asynchronously (for example, by passing the `cloudServices` config), you may observe that sometimes it falls into an infinite loop. I've checked, and it seems to be connected with the `_destroyEditor` function.

Let's check and fix it in this PR since it is connected to the same issue.

DawidKossowski (approved):

glynam1:
@Mati365 thanks for your work on this. Did you get a change to test the context editor for this? Just wondering should I look to upgrade when this gets released.
Thanks

Mati365:
@glynam1 I checked our internal test suite and it worked fine. imo it might really depend on your testing library set, so I'm not sure if my current solution solves your issue. You have to check.

glynam1:
Cool thanks, I'll open a ticket if I see the error reoccurring after this upgrade.

PR #490, opened 2024-07-04
Migrate to ESM

filipsobol:

Internal: Migrate from webpack and Karma to Vite and Vitest. See ckeditor/ckeditor5#16616.

Internal: Fix test coverage which only covered a small portion of the codebase.

MINOR BREAKING CHANGE: Migrate to ESM. Related to ckeditor/ckeditor5#16616.

---

### Additional information


src/demos/multiroot-react/main.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const element = document.getElementById( 'root' ) as HTMLDivElement;
+
+if ( __REACT_VERSION__ === 16 ) {
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-ignore
+	const ReactDOM = await import( 'react-dom' );
+
+	ReactDOM.render( React.createElement( App ), element );
+} else {
```
Mati365, line 17: 👍🏻 Nice idea to use compiler global to decide which react to choose.

filipsobol:
Failing tests are related to our migration from Enzyme to React Testing Library. They will be fixed in other PR and merged into this branch.

src/demos/react/ContextDemo.tsx
```diff
@@ -33,7 +45,6 @@ export default function ContextDemo( props: ContextDemoProps ): JSX.Element {
 			<h2 className="subtitle">Editor Context Demo</h2>
 			<p className="info">Component&apos;s events are logged to the console.</p>
 
-			{ /* @ts-expect-error: Caused by linking to parent project and conflicting react types */ }
 			<CKEditorContext
```
Mati365, line 48: Cool, these errors frustrating. It's nice that they no longer occur.

Mati365 (commented):
I take a look briefly and it looks nice. During migration of the tests I'll double-check these changes and if nothing explodes then we can assume that everything is ok ;)

tests/_utils/render.tsx
```diff
@@ -0,0 +1,27 @@
+/**
```
filipsobol, line 1: Rename file to `promisemanager.ts`

Mati365 (approved):
I checked it during development of tests, and it seems to be working correctly 🎉

package.json
```diff
@@ -76,16 +75,19 @@
     "node": ">=18.0.0"
   },
   "scripts": {
+    "dev": "echo \"Use 'dev:16', 'dev:18', or 'dev:19' depending on the version of React you want to test\"",
+    "dev:16": "REACT_VERSION=16 vite",
+    "dev:18": "REACT_VERSION=18 vite",
+    "dev:19": "REACT_VERSION=19 vite",
```
martnpaneq, line 81: I like the new way of testing different react versions. What do you think about making the latest version to be the default?

tests/_utils-tests/context.test.tsx
```diff
@@ -17,7 +18,7 @@ describe( 'Context', () => {
 
 	describe( 'destroy()', () => {
 		it( 'should return a promise that resolves properly', () => {
-			return Context.create()
+			return Context.create( )
```
martnpaneq, line 21: Space got here by mistake?

scripts/preparepackages.js
```diff
@@ -9,9 +9,9 @@
 
 'use strict';
 
-const { Listr } = require( 'listr2' );
-const releaseTools = require( '@ckeditor/ckeditor5-dev-release-tools' );
-const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
+import { Listr } from 'listr2';
+import releaseTools from '@ckeditor/ckeditor5-dev-release-tools';
+import { tools } from '@ckeditor/ckeditor5-dev-utils';
```
martnpaneq, line 14: We are using `require` in this package in line 55. Please make sure that this works well with modules. 
image

martnpaneq:
@filipsobol @pomek That's a very nice cleanup of the repo and I like that you managed to get rid of `demos` directories. Please have a look at the minor comments that I left and this should be ready to merge :)

Reinmar:
Why are we removing all the demos? Are we sure they didn't serve some purpose?

At least one of them is frequently visited.

Mati365:
@Reinmar demos are still there

Reinmar:
> @Reinmar demos are still there

Where?

:point\_up: this was a an honest question, but I realized that since you asked they are probably somewhere...

I would never think to look for them in `src/demos/` :open\_mouth: And I don't get the purpose of `./index.html`. Who and how is supposed to open that file in the browser? Is this a common pattern among React libraries?

Mati365:
@Reinmar I get your point. Yep, `src/` placement might be misleading. It's not the biggest problem. I'm sure that there will be accidential imports between manuał/ and production code which we don't want for sure.

Reinmar:
I talked about this with @filipsobol too.

We also saw the need to update https://github.com/ckeditor/ckeditor5-react/tree/master?tab=readme-ov-file#contributing

package.json
Reinmar: This may decrease visibility of other options.

Those demos have quite unclear audience. I think that mostly us are running them. Other people, to my intuition, probably just scan their code. 

Exposing the ability to test with previous React versions to us would be a nice benefit. It increases the awareness of potential issues.

filipsobol:
TODO: Check Vite `target`

src/ckeditor.tsx
```diff
@@ -254,7 +254,7 @@ export default class CKEditor<TEditor extends Editor> extends React.Component<Pr
 
 			const instance = await this._createEditor( el as any, config );
 
-			if ( totalRestartsRef.current > 0 ) {
+			if ( editorSemaphore && totalRestartsRef.current > 0 ) {
```
Reinmar, line 257: Why are we changing this? Is it part of this PR?

filipsobol: @Mati365

filipsobol:
I addresses all comments and suggestions. The demos were moved from `/src/demos/*` to `/demos/*` so they are more visible.

src/ckeditor.tsx
Reinmar: I asked whether this is part of this PR's scope?

If not, it should be extracted. As I understand, this is a bonus change. Please open a ticket for it (it will also explain why it's needed because I still don't know) and move it there.

Mati365: Ah, I got it. Sorry. Yep, you are correct. I checked it a bit more, and it's leftover after writing tests in this region. Nullability check might be skipped there, restart of the editor occurs only if semaphore is present. Please reopen this thread in case of further comments.

Mati365:
fyi, I discovered one issue related to incorrect detection of Context in `useMultiRootEditor` hook. Long story short - our enzyme tests were not testing context forwarding at all, and it was possible to use the wrong context while having the test suite succeed. That was introduced in [React 19 integration fixes](https://github.com/ckeditor/ckeditor5-react/commit/eaccdab2619551fbc867e87cac4b78e1ebab59ee), and it was not released. It's not a huge issue, and it was much easier to do that in the scope of this PR due to lack of proper enzyme testing.

https://github.com/ckeditor/ckeditor5-react/pull/490/files#diff-a495913bafb3d316e8cc54a39c3de3a4f04b1746ee0f40a0e5dd722d0a955533R368-R375

src/useMultiRootEditor.tsx
```diff
@@ -241,7 +244,7 @@ const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookReturns =
 	/**
 	 * Callback function for handling a detached root.
 	 */
-	const onDetachRoot = useRefSafeCallback( ( editor: MultiRootEditor, evt: EventInfo, root: RootElement ): void => {
+	const onDetachRoot = useRefSafeCallback( ( _: MultiRootEditor, __: EventInfo, root: RootElement ): void => {
```
Reinmar, line 247: Why was this change needed?

filipsobol: This is a TypeScript thing. It shows warnings when there are unused function parameters and the way to remove these warnings is to rename the parameters to start with `_`. Common convention is to use `_` when you don't care about given parameter, but for readability I renamed them to `_editor` and `_evt`.

package.json
```diff
@@ -25,67 +31,61 @@
...
   "engines": {
     "node": ">=18.0.0"
   },
   "scripts": {
+    "dev": "echo \"Use 'dev:16', 'dev:18', or 'dev:19' depending on the version of React you want to test\"",
+    "dev:16": "REACT_VERSION=16 vite",
+    "dev:18": "REACT_VERSION=18 vite",
+    "dev:19": "REACT_VERSION=19 vite",
+    "build": "vite build && tsc --emitDeclarationOnly",
+    "test": "vitest run --coverage",
```
DawidKossowski, line 82: As I understand it correctly, the tests will be processed with React 18 by default. Should we provide similar scripts for other React versions and one main command that will process tests using the all main supported React versions?

DawidKossowski (commented):
I can see that you changed the convention of file names in tests. Why did you make that change? Did you document this decision in any logbook?

tsconfig.json
```diff
@@ -1,24 +1,32 @@
 {
   "compilerOptions": {
+    "target": "ES2020",
```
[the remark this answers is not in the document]
filipsobol: Nice catch. I updated `target` and `lib` to match TypeScript config we used in CKE5 repo.

package.json
Mati365: @DawidKossowski I tried to do that, but there is versioning issue. Latest React Testing Library requires minimum React 18, and does not work on 16. On the other hand, older RTL does not support React 19.

filipsobol: We've tried that, but it's more difficult than running the development environment with different versions of React.

Different versions of React have slightly different APIs for starting it. While in demos we only call these APIs in two places where we can use simple `if` statements, in tests we do a lot more of it.
Also, in the development environment, React is the only thing we need to swap. But in tests, it's also different versions of the React testing library, which makes it even more tricky.

It's not impossible, but it would add a lot of complexity.

tsconfig.json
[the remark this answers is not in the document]
DawidKossowski: @filipsobol, can you explain why we can use optional chaining (and other) operators in this repo but cannot use them in the CKE5 repo even though both repos support es2019?

filipsobol:
As discussed, I added the UMD build for better backward compatibility, following the configuration recommendations from the Vite documentation: https://vitejs.dev/guide/build#library-mode.

tsconfig.json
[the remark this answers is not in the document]
filipsobol: In both the CKE5 repo and here we **could** use optional chaining, since in both environments our target is `ES2019`. With such a target, a simple `test?.prop1?.prop2` is transformed into `(_a = test === null || test === void 0 ? void 0 : test.prop1) === null || _a === void 0 ? void 0 : _a.prop2` (or similar, depending on the tooling), which works well in older environments like webpack 4, which we still support in the old installation methods.

However, in CKE5 the code is transformed BEFORE the tests are run, meaning that the code coverage is generated for the transformed code instead of simple `test?.prop1?.prop2`. This messes up the code coverage and generally makes it not worth using.

We don't have this problem in this repository, so we can use it just fine.

DawidKossowski (approved):

PR #498, opened 2024-07-10
Release the repository from CI

pomek:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Internal: Redesign the release process to allow it to be processed from CircleCI.

---

### Additional information

* Part of https://github.com/cksource/ckeditor5-internal/issues/3710.
* Requires: https://github.com/ckeditor/ckeditor5-dev/pull/969.

PR #501, opened 2024-07-12
Remove top-level await from the `postinstall` script

filipsobol:

Internal: Remove top-level await from the `postinstall` script.

---

### Additional information


pomek (approved):

pomek (commented):
I gave R+ when it should be R-.

Now I remember why we decided to use a dynamic import. Please, see https://github.com/ckeditor/ckeditor5-react/commit/032f745260cbfc34bcb6ffb83ee66142af6b93e2.

pomek:
I will handle it #498.

PR #506, opened 2024-07-17
Fix: Change JSX runtime to `classic`.

filipsobol:

Fix: Change JSX runtime to `classic` so that `react-jsx-runtime` is not bundled.

---

### Additional information


Mati365:
Tested locally, works good!

Mati365 (approved):

pomek (approved):

Issue #513, opened 2024-08-09
CkeditorContext onReady function called before editors attach to context


glynam1:
This seems to have been introduced in 7.0.0

When using the ckeditor react context demo, attempt to log the editors after onReady is called:

```
			<CKEditorContext
				context={ ClassicEditor.Context as any }
				contextWatchdog={ ClassicEditor.ContextWatchdog as any }
				onReady={context => {
					console.log( context.editors.length );
				}}
			>
```
			
This will log out as 0.

Based on these findings onReady is being called too early, as functionality that relies on this behaviour to populate the list of editors in the context will break. 

This essentially breaks our setup and prevents us from upgrading, as bar listening on every <CkEditor> onReady callback, and then populating a list based on that, there is no way to populate this

Mati365:
Hi! Thanks for report. I'll check this.

Mati365:
@glynam1 Good catch! It looks like our react integration docs are a bit outdated and we'll update them. Basically `onReady` behavior has been changed after fixing race conditions in our integration introduced with newest React versions.

At this moment `onReady` is called after initialization of context and after that we attach editors to it. So if you want to store editors in some kind of state you can do:

```tsx
import type { CollectionChangeEvent, Editor } from 'ckeditor5';

..

<CKEditorContext
  context={ ClassicEditor.Context as any }
  contextWatchdog={ ClassicEditor.ContextWatchdog as any }
  onReady={ context => {
    context.editors.on<CollectionChangeEvent<Editor>>( 'change', ( ev, data ) => {
      console.info( data.added, data.removed );
      // or
      console.info( context.editors );
    } );
  } }
>
 ..
</CKEditorContext>
```

Mati365:
I'm closing this issue, please reopen it if the problem still occurs.

glynam1:
@Mati365 does an event emit when the context has reached a steady state? It seems like this is a change of behaviour, as onReady used to fire at the end of startup, whereas with the above snippet, the "end state" is unknown

Mati365:
@glynam1

> does an event emit when the context has reached a steady state

Yep, but it does not mean that context has assigned any editor. 

> It seems like this is a change of behaviour, as onReady used to fire at the end of startup, whereas with the above snippet, the "end state" is unknown

[...]

PR #514, opened 2024-08-09
Add `onTrackInitializedEditors` callback to `CKEditorContext`

Mati365:

Feature: Add `onTrackInitializedEditors` callback to `CKEditorContext`. Closes https://github.com/ckeditor/ckeditor5-react/issues/513

---

### Additional information

During React 19 refactor, it was discovered that `CKEditorContext` must be initialized _before_ rendering the rest of React tree. It was because in the `StrictMode` the `CKEditorContext` was recreated in small amount of the time which was causing crashes in nested editors (which were using wrong reference of `CKEditorContext` during initialization).

This PR adds `onTrackInitializedEditors` callback to  `CKEditorContext` that is fired after any initialization or destruction of editor.

### DIFF
--- a/demos/react/ContextDemo.tsx
+++ b/demos/react/ContextDemo.tsx
@@ -48,6 +48,9 @@ export default function ContextDemo( props: ContextDemoProps ): JSX.Element {
 			<CKEditorContext
 				context={ ClassicEditor.Context as any }
 				contextWatchdog={ ClassicEditor.ContextWatchdog as any }
+				onTrackInitializedEditors={ editors => {
+					console.log( 'Editors:', editors );
+				}}
 			>
 				<div className="buttons">
 					<button
@@ -59,6 +62,7 @@ export default function ContextDemo( props: ContextDemoProps ): JSX.Element {
 				</div>
 
 				<CKEditor
+					id='abc'
 					editor={ ClassicEditor as any }
 					data={ props.content }
 					onReady={ ( editor: any ) => {
--- a/src/ckeditor.tsx
+++ b/src/ckeditor.tsx
@@ -27,7 +27,7 @@ import {
 	ContextWatchdogContext,
 	isContextWatchdogInitializing,
 	isContextWatchdogReadyToUse
-} from './ckeditorcontext';
+} from './context/ckeditorcontext';
 
 const REACT_INTEGRATION_READ_ONLY_LOCK_ID = 'Lock from React integration (@ckeditor/ckeditor5-react)';
 
--- a/src/context/useInitializedCKEditorsMap.ts
+++ b/src/context/useInitializedCKEditorsMap.ts
@@ -1,0 +1,102 @@
+/**
+ * @license Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import { useEffect } from 'react';
+import { useRefSafeCallback } from '../hooks/useRefSafeCallback';
+
+import type { CollectionAddEvent, Context, Editor } from 'ckeditor5';
+import type { ContextWatchdogValue } from './ckeditorcontext';
+
+/**
+ * A hook that listens for the editor initialization and destruction events and updates the editors map.
+ *
+ * @param config The configuration of the hook.
+ * @param config.currentContextWatchdog The current context watchdog value.
+ * @param config.onChangeEditorsMap The function that updates the editors map.
+ * @example
+ * ```ts
+ * useInitializedCKEditorsMap( {
+ * 	currentContextWatchdog,
+ * 	onTrackInitializedEditors: ( editors, context ) => {
+ * 		console.log( 'Editors:', editors );
+ * 	}
+ * } );
+ * ```
+ */
+export const useInitializedCKEditorsMap = <TContext extends Context>(
+	{
+		currentContextWatchdog,
+		onTrackInitializedEditors
+	}: InitializedContextEditorsConfig<TContext>
+): void => {
+	// We need to use the safe callback to prevent the stale closure problem.
+	const onTrackInitializedEditorsSafe = useRefSafeCallback( onTrackInitializedEditors || ( () => {} ) );
+
+	useEffect( () => {
+		if ( currentContextWatchdog.status !== 'initialized' ) {
+			return;
+		}
+
+		const { watchdog } = currentContextWatchdog;
+		const editors = watchdog?.context?.editors;
+
+		if ( !editors ) {
+			return;
+		}
+
+		// Get the initialized editors from
+		const getInitializedContextEditors = () => [ ...editors ].reduce(
+			( map, editor ) => {
+				if ( editor.state === 'ready' ) {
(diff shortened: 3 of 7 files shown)


src/context/useInitializedCKEditorsMap.ts
```diff
@@ -0,0 +1,118 @@
...
+
+		// Get the initialized editors from
+		const getInitializedContextEditors = () => [ ...editors ].reduce<InitializedEditorsMap>(
+			( map, editor ) => {
+				if ( editor.state !== 'ready' ) {
+					return map;
+				}
+
+				const metadata = tryExtractCKEditorReactContextMetadata( editor.config );
+				const nameOrId = metadata ? metadata.editorName : editor.id;
```
filipsobol, line 62:
```suggestion
				const nameOrId = metadata?.editorName ?? editor.id;
```

Nitpick

src/ckeditor.tsx
```diff
@@ -440,6 +452,7 @@ export interface Props<TEditor extends Editor> extends InferProps<typeof CKEdito
 		EditorWatchdog: typeof EditorWatchdog;
 		ContextWatchdog: typeof ContextWatchdog;
 	};
+	context?: CKEditorConfigContextMetadata;
```
filipsobol, line 455: What is the purpose of introducing a new property that accepts arbitrary data?

I feel like this integration already deals with too many things called `context` and a new one will only make things worse. Wouldn't adding a new optional property called `id` or `name` be sufficient and easier to document?

demos/react/ContextDemo.tsx
```diff
@@ -48,43 +40,42 @@ export default function ContextDemo( props: ContextDemoProps ): JSX.Element {
 			<CKEditorContext
 				context={ ClassicEditor.Context as any }
 				contextWatchdog={ ClassicEditor.ContextWatchdog as any }
+				onChangeEditorsMap={ editors => {
```
filipsobol, line 43: Naming things is a matter of personal taste, but for me, the name `onChangeEditorsMap` doesn't properly describe what this prop does. How about changing it to something more descriptive like `onEditorsStateChange` or similar?

filipsobol (commented):
I'm not that good with React, so I will let others do the proper review, but I left some thoughts about the public API.

demos/react/ContextDemo.tsx
Mati365: I'm not super fan of `onChangeEditorsMap` too, but I'm not sure if `onEditorsStateChange` is super clear too imho. The problem is that `onEditorsStateChange` would tell user that it tracks transitions between loading and loaded state, which is not the case in this callback. It tracks only initialized / destroyed editors.

src/ckeditor.tsx
Mati365: I was struggling with this too. `id` cannot be used because it is used to decide if the editor should be reloaded or not. So I'm nearly 100% sure that some of the integrations do something like this: `id={JSON.stringify(config)}` which is not we want in editors, mam. The `name` on the other hand can be misleading with HTML `name` attribute. The main concept of `context` property is to:

1. Tell the user that it's only context-only configuration, so it'll be ignored if the component is not used within context.
2. Pass more additional fields from child to parent. In other words, we can add some metadata about placement of the editor to context parent. Like this:

```tsx
<CKEditorContext
  context={ ClassicEditor.Context }
  contextWatchdog={ ClassicEditor.ContextWatchdog }
  onChangeEditorsMap={ editors => {
    console.log( editors.editor2.stuff );
  }}
>
  <CKEditor
    editor={ ClassicEditor }
    data="<h2>Another Editor</h2><p>... in common Context</p>"
    context={{
      editorName: 'editor2',
      stuff: 2
    }}
  />
</CKEditorContext>
```

`Stuff` is forwarded to parent and can be used to forward some `CKEditor` state to parent using our context.

demos/react/ContextDemo.tsx
filipsobol: Other suggestions based on our conversation:

* `onEditorsUpdate`, but it might suggest that it gets called whenever editor content changes
* `onEditorsLifecycleUpdate`
* `onChangeEditorsList`

glynam1:
Hi,

Just a question on this implementation so I can confirm closure of https://github.com/ckeditor/ckeditor5-react/issues/513 when this is merged, `onChangeEditorsMap` will be fired when an editor is ready right? 
So this is equivalent to listening to the onReady callback from an editor.

If this is the same as the callback from 
`context.editors.on<CollectionChangeEvent<Editor>>( 'change', ( ev, data ) => {`

Then I think this won't be useful as the editor is still not actually ready.

Mati365:
@glynam1 It'll be fired when editor is ready (and after destroying of editor). It's the flat map of fully initialized editors that are present in React tree.

glynam1:
> @glynam1 It'll be fired when editor is ready (and after destroying of editor). It's the flat map of fully initialized editors that are present in React tree.

Nice one, looking forward to the update so!

demos/react/ContextDemo.tsx
arkflpc: `onChangeEditors`? What is the meaning of `Map` suffix?

arkflpc:
LGTM, but the naming is not clear enough.

demos/react/ContextDemo.tsx
Mati365: @arkflpc All editors are stored in context tree using dictionary: `name -> { instance: ... }` format.

```tsx
{
  editor1: { instance: ... },
  editor2: { instance: ... }
}
```

Mati365: `onWatchInitializedEditors` maybe more descriptive, wdyt @arkflpc ?

Mati365:
@arkflpc I renamed callback to `onWatchInitializedEditors`. Can you take a look?

demos/react/ContextDemo.tsx
gorzelinski: Tbh, I'm also not a fan of this `map` suffix. Maybe that's just me, but I would avoid generic nouns or data types in such names. Your alternative is good. But I also think the `onChange` convention is pretty well known. So, I thought about other options, something like `onChangeInitializedEditors`.

gorzelinski (commented):
I spotted one thing.

demos/react/ContextDemo.tsx
```diff
@@ -48,43 +40,42 @@ export default function ContextDemo( props: ContextDemoProps ): JSX.Element {
...
 						disabled={ !state.editor1 }
 					>
 						Simulate an error in the first editor
 					</button>
 				</div>
 
 				<CKEditor
+					context={{
+						editorName: 'editor1'
+					}}
```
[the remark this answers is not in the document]
Mati365: Good catch, thx!

demos/react/ContextDemo.tsx
Mati365: Renamed, imho good enough. It'll be documented anyway.

gorzelinski (approved):

PR #535, opened 2024-09-24
Fix: Update data of roots with modified content only

### LINKED ISSUE #534: Setting data of any root with MultiRoot editor results in revision history marking all roots as changed
## Reproduction steps

_It can be done in multiple ways, the fastest for me was by using collaboration samples._

1. Checkout collaboration samples repo and follow setup instructions.
2. In [real-time-collaboration-editor-multi-root-for-react](https://github.com/ckeditor/ckeditor5-collaboration-samples/tree/master/real-time-collaboration-editor-multi-root-for-react) expose `data` and `setData` returned by `useMultiRootEditor`. This can be done by modifying [this line](https://github.com/ckeditor/ckeditor5-collaboration-samples/blob/acf29d8194ef8929cf93ab4317ea81c17c846b1d/real-time-collaboration-editor-multi-root-for-react/src/App.jsx#L528) as:

```ts
	const { data, setData, editableElements, toolbarElement, attributes } = useMultiRootEditor( editorProps );

	window.setData = setData;
	window.data = data;
```

3. Run the sample with `yarn && yarn dev`.
4. Go to http://localhost:5173/.
5. Open dev console, paste and execute `window.setData({...window.data, header: '123123'})`.
6. Go to editor revision history.

### Expected

Only `header` root is marked as changed.

### Actual

All roots are marked as changed:

image

## A bit of context

This happens because `setData` returned from `useMultiRootEditor` hook, expects all roots to be passed (since it diffs root list enabling to add/remove them) and existing roots are just updated with passed data, even if it's the same. Such update results in each root model data to be wiped out and re-added which is interpreted by revision history as complete data replace (since it works on model operations and does not diff data changes).

f1ames:

Fix: Update data of roots with modified content only. Closes #534.

---

### Additional information

See #534 for more details. TL;DR:

> This happens because `setData` returned from `useMultiRootEditor` hook, expects all roots to be passed (since it diffs root list enabling to add/remove them) and existing roots are just updated with passed data, even if it's the same. Such update results in each root model data to be wiped out and re-added which is interpreted by revision history as complete data replace (since it works on model operations and does not diff data changes).

This PR slightly modifies the logic to update only the roots which has changed data.

---

I dug a bit in the context of below comment in the code:

> // If any of the roots content has changed, set the editor data.
> // Unfortunately, we cannot set the editor data just for one root,
> // so we need to overwrite all roots (`nextProps.data` is an
> // object with data for each root).

But didn't find any case or code which would suggest that we can't do it. All tests green, manuals works fine AFAIT too.

### DIFF
--- a/src/useMultiRootEditor.tsx
+++ b/src/useMultiRootEditor.tsx
@@ -505,7 +505,7 @@ const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookReturns =
 				data || /* istanbul ignore next -- @preserve: It should never happen, data should be always filled. */ {}
 			);
 
-			const hasModifiedData = dataKeys.filter( rootName =>
+			const modifiedRoots = dataKeys.filter( rootName =>
 				editorData[ rootName ] !== undefined &&
 				JSON.stringify( editorData[ rootName ] ) !== JSON.stringify( data[ rootName ] )
 			);
@@ -531,7 +531,10 @@ const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookReturns =
 			};
 
 			const _updateEditorData = ( roots: Array<string> ) => {
-				const dataToUpdate = roots.reduce( ( result, rootName ) => ( { ...result, [ rootName ]: data[ rootName ] } ), {} );
+				const dataToUpdate = roots.reduce(
+					( result, rootName ) => ( { ...result, [ rootName ]: data[ rootName ] } ),
+					Object.create( null )
+				);
 				instance.data.set( dataToUpdate, { suppressErrorInCollaboration: true } as any );
 			};
 
@@ -552,8 +555,8 @@ const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookReturns =
 					_handleNewRoots( newRoots );
 					_handleRemovedRoots( removedRoots );
 
-					if ( hasModifiedData.length ) {
-						_updateEditorData( hasModifiedData );
+					if ( modifiedRoots.length ) {
+						_updateEditorData( modifiedRoots );
 					}
 
 					if ( rootsWithChangedAttributes.length ) {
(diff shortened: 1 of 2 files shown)


src/useMultiRootEditor.tsx
```diff
@@ -530,12 +530,9 @@ const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookReturns =
...
 
-			const _updateEditorData = () => {
-				// If any of the roots content has changed, set the editor data.
-				// Unfortunately, we cannot set the editor data just for one root,
-				// so we need to overwrite all roots (`nextProps.data` is an
-				// object with data for each root).
-				instance.data.set( data, { suppressErrorInCollaboration: true } as any );
+			const _updateEditorData = ( roots: Array<string> ) => {
+				const dataToUpdate = roots.reduce( ( result, rootName ) => ( { ...result, [ rootName ]: data[ rootName ] } ), {} );
+				instance.data.set( dataToUpdate, { suppressErrorInCollaboration: true } as any );
```
Mati365, line 535: Consider using `Object.create( null )` as initial value of reduce.

src/useMultiRootEditor.tsx
```diff
@@ -505,7 +505,7 @@ const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookReturns =
 				data || /* istanbul ignore next -- @preserve: It should never happen, data should be always filled. */ {}
 			);
 
-			const hasModifiedData = dataKeys.some( rootName =>
+			const hasModifiedData = dataKeys.filter( rootName =>
```
DawidKossowski, line 508: Please rename this variable as it is no longer a boolean, but an array of strings.

DawidKossowski (changes requested):

src/useMultiRootEditor.tsx
f1ames: Fixed in f09915b7593bdbb57cca4e04221b2088fb69e43f.

src/useMultiRootEditor.tsx
f1ames: Good point, fixed in f09915b7593bdbb57cca4e04221b2088fb69e43f.

Mati365 (approved):

DawidKossowski (approved):

PR #541, opened 2024-09-30
Fix inconsistent behavior of `useAsyncValue` and `useCKEditorCloud` hooks on `Vite` and `Next` runtimes.

Mati365:

Fix: Behavior of `useCKEditorCloud` hook is now consistent on `Vite` and `Next` runtimes while changing properties.
Internal: Fix incorrect typings of `useAsyncCallback` and `useAsyncValue` callback arguments types.
Tests: Add missing `cleanup` methods to few tests.

---

### Additional information

It looks like `useRef` behavior is inconsistent across React frameworks, and it is not related to React version itself (at least 18.x one, 19.x is not affected at all).

Let's assume this component that is being rendered in strict mode:

```tsx
const Component = () => {
    const ref = useRef(null);

    console.info(ref.current);

    ref.current = 'test';

    return null;
};
```

1. On **Vite** it will print `null` two times.
2. On **Next.js** it will print `null` the first time, and then `test` second time.

It looks like something changed somewhere in React world and second approach is no longer consistent across frameworks. At the time of creating `useInstantEffect` it was. It affects our `useCKEditorCloud` which uses `useInstantEffect` under the hood, which may work differently across frameworks above. This PR normalizes it.

### DIFF
--- a/src/hooks/useInstantEffect.ts
+++ b/src/hooks/useInstantEffect.ts
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md.
  */
 
-import { useRef, type DependencyList } from 'react';
+import { useState, type DependencyList } from 'react';
 import { shallowCompareArrays } from '@ckeditor/ckeditor5-integrations-common';
 
 /**
@@ -13,10 +13,10 @@ import { shallowCompareArrays } from '@ckeditor/ckeditor5-integrations-common';
  * @param deps The dependency list.
  */
 export const useInstantEffect = ( fn: VoidFunction, deps: DependencyList ): void => {
-	const prevDeps = useRef<any>( null );
+	const [ prevDeps, setDeps ] = useState<any>( null );
 
-	if ( !shallowCompareArrays( prevDeps.current, deps ) ) {
-		prevDeps.current = [ ...deps ];
+	if ( !shallowCompareArrays( prevDeps, deps ) ) {
 		fn();
+		setDeps( [ ...deps ] );
 	}
 };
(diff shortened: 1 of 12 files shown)


filipsobol (approved):

filipsobol:
LGTM, but as I have little knowledge of React, I'd be good if someone else looked at this PR too.

src/hooks/useInstantEffect.ts
```diff
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md.
```
gorzelinski, line 3: I found a problem with our es rule - it's probably not directly related to the PR, but it's worth noting. Every file complains that the rule about the license key cannot be found: `Definition for rule 'ckeditor5-rules/prevent-license-key-leak' was not found.`

gorzelinski (commented):
I dug a little deeper. The inconsistencies may occur because of the caveats with the refs. TLDR version is this - do not write or read `ref.current` during rendering. If you have to read or write something during rendering, use state instead. So, the solution is probably ok. The snippet below works consistently across frameworks:

```jsx
'use client'
import { useEffect, useRef } from "react";

export const Component = () => {
    const ref = useRef(null);

    console.info(ref.current);

    useEffect(() => {
        // ✅ You can read or write refs in effects
        ref.current = 'test';
    });

    return null;
};
```

I also added a comment that's probably unrelated.

src/hooks/useInstantEffect.ts
Mati365: fyi @filipsobol

Mati365:
Thanks for digging it more @gorzelinski, I'm merging then.

Issue #542, opened 2024-10-02
Flaky 'useMultiRootEditor' slow editor test


f1ames:
The `should assign properly data property to editor even if it is still mounting` (see [here](https://github.com/ckeditor/ckeditor5-react/blob/8886ca37769cd5145e6c6249ea5859a60ee9120b/tests/useMultiRootEditor.test.tsx#L903)) test fails randomly with:

```
- Expected
+ Received

  Object {
    "content": "",
-   "intro": "Hello World!",
+   "intro": "",
  }

 ❯ tests/useMultiRootEditor.test.tsx:951:57
    949|     await waitFor( () => {
    950|      expect( result.current.editor ).to.be.instanceof( SlowEditor );
    951|      expect( result.current.editor!.data.get() ).to.deep.equal( {
       |                                                         ^
    952|       intro: 'Hello World!',
    953|       content: ''
```
_See [on CI](https://app.circleci.com/pipelines/github/ckeditor/ckeditor5-react/889/workflows/0369ca63-9b10-41bc-8c00-21a763f7bc92/jobs/1565)_.

https://github.com/ckeditor/ckeditor5-react/blob/8886ca37769cd5145e6c6249ea5859a60ee9120b/tests/useMultiRootEditor.test.tsx#L903

It might be related to using timeouts:

https://github.com/ckeditor/ckeditor5-react/blob/8886ca37769cd5145e6c6249ea5859a60ee9120b/tests/useMultiRootEditor.test.tsx#L938-L955

What's worth mentioning is that it seems it got more unstable after #534 PR got merged.

f1ames:
After #543, it seems this is not flaky test in fact, but really a regression from #534. Test is failing randomly still after improvements from #543 so there might be race condition or something similar going on.

f1ames:
Ok, some initial observations. I'll put my console dump first (you can skip it first and go to conclusions first) which might not be really readable but still (and at least useful for me) and then refer to it.

You can find a diff of the above here - https://www.diffchecker.com/H91iddhW/.

### Successful run

```
----- START -----
useMultiRootEditor:return - editorRefs.instance.current?.id undefined
useMultiRootEditor:return - editorRefs.instance.current?.id undefined
useLifeCycleSemaphoreSyncRef:runAfterMount - revision 1728908958656
useMultiRootEditor:_initializeEditor - totalRestartsRef {
  "current": 0,
}
[...]

PR #546, opened 2024-10-15
Restore restarted watchdog editor with current react state

Mati365:

Fix: Prevent potential crashes of `useMultiRootEditor` during the initialization phase when setting the new state of the multi-root editor with an attached watchdog. Closes https://github.com/ckeditor/ckeditor5-react/issues/542

---

### Additional information

It appears that some of our tests are causing random restarts of the editor. An example of such a restart might occur when setting some data while the component has not fully rendered the entire CKEditor, and the semaphore begins destroying it while running in the watchdog context.

This PR resets the reinitialized editor data to the one defined in the React state (data is defined in the upper scope). This change seems to make our tests more stable and should be harmless.

### DIFF
--- a/src/useMultiRootEditor.tsx
+++ b/src/useMultiRootEditor.tsx
@@ -380,7 +380,13 @@ const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookReturns =
 			current: 0
 		};
 
-		watchdog.setCreator( async ( data, config ) => {
+		// Keeping using `data` from creator function callback seems to be a good idea in theory,
+		// but in practice, it leads to instability. The `data` object can be changed during the editor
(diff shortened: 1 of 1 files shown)


f1ames (changes requested):
Should we have a dedicated test for that? Now we have one which sometimes fails due to previous behavior (but it may not), so this specific logic is not really covered 🤔

Mati365:
@f1ames I don't think so, simulating this is too difficult, and it's that kind of behavior that often changes with React implementation changes.

f1ames (approved):
> @f1ames I don't think so, simulating this is too difficult, and it's that kind of behavior that often changes with React implementation changes.

Ok, I agree. LGTM then 👍

PR #559, opened 2024-11-13
Add retries to flaky multiroot semaphores tests.

Mati365:

Tests: Add `retry` option to flaky `multiroot` test, extend delay between assertions. 
Tests: Randomize the order of the tests, to prevent incorrect dependencies across tests. 

---

### Additional information

This test tests quite rare scenario when `setData` is being called when editor is not initialized. In theory - it is possible, but I haven't managed to reproduce this in the real world. At this moment, there were no reports about incorrect behavior, so it's safe enough to retry it a few times.

### DIFF
--- a/tests/useMultiRootEditor.test.tsx
+++ b/tests/useMultiRootEditor.test.tsx
@@ -931,7 +931,7 @@ describe( 'useMultiRootEditor', () => {
 					editor: SlowEditor
 				} ) );
 
-				await timeout( 100 );
+				await timeout( 500 );
 
 				// Depending on the execution order on the event loop this `setData` might be delayed by the React engine.
 				// It happens only if event loop is busy and React has to wait for the next tick a little bit longer than usual.
@@ -942,9 +942,9 @@ describe( 'useMultiRootEditor', () => {
 					content: ''
 				} );
 
-				await timeout( 200 );
+				await timeout( 500 );
 
-				deferInitialization.resolve();
+				await deferInitialization.resolve();
 
 				await waitFor( () => {
 					expect( result.current.editor ).to.be.instanceof( SlowEditor );
--- a/tests/useMultiRootEditor.test.tsx
+++ b/tests/useMultiRootEditor.test.tsx
@@ -896,7 +896,7 @@ describe( 'useMultiRootEditor', () => {
 
 	describe( 'semaphores', () => {
 		const testSemaphoreForWatchdog = enableWatchdog => {
-			it( 'should assign properly `data` property to editor even if it is still mounting', { retry: 3 }, async () => {
+			it( 'should assign `data` property to the editor even if it is still mounting', { retry: 3 }, async () => {
 				const deferInitialization = createDefer();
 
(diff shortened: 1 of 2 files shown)


tests/useMultiRootEditor.test.tsx
```diff
@@ -896,7 +896,7 @@ describe( 'useMultiRootEditor', () => {
 
 	describe( 'semaphores', () => {
 		const testSemaphoreForWatchdog = enableWatchdog => {
-			it( 'should assign properly `data` property to editor even if it is still mounting', async () => {
+			it( 'should assign properly `data` property to editor even if it is still mounting', { retry: 3 }, async () => {
```
gorzelinski, line 899:
```suggestion
			it( 'should assign `data` property to the editor even if it is still mounting', { retry: 3 }, async () => {
```

tests/useMultiRootEditor.test.tsx
```diff
@@ -931,16 +931,20 @@ describe( 'useMultiRootEditor', () => {
 					editor: SlowEditor
 				} ) );
 
-				await timeout( 100 );
+				await timeout( 500 );
 
+				// Depending on the execution order on the event loop this `setData` might be delayed by the React engine.
```
gorzelinski, line 936:
```suggestion
				// Depending on the execution order on the event loop, this `setData` might be delayed by the React engine.
```

tests/useMultiRootEditor.test.tsx
```diff
@@ -931,16 +931,20 @@ describe( 'useMultiRootEditor', () => {
...
+				// It happens only if event loop is busy and React has to wait for the next tick a little bit longer than usual.
```
gorzelinski, line 937:
```suggestion
				// It happens only if event loop is busy and React has to wait for the next tick a little bit longer than usual.
```
```suggestion
				// It happens only if the event loop is busy and React has to wait for the next tick a little bit longer than usual.
```

tests/useMultiRootEditor.test.tsx
```diff
@@ -931,16 +931,20 @@ describe( 'useMultiRootEditor', () => {
...
+				// It does not play well with `waitFor` below so we added few retries to make it more stable.
```
gorzelinski, line 938:
```suggestion
				// It does not play well with the `waitFor` below, so we added a few retries to make it more stable.
```

tests/useMultiRootEditor.test.tsx
```diff
@@ -931,16 +931,20 @@ describe( 'useMultiRootEditor', () => {
...
+				// It should not be a problem in real life, as it is a rare case and might be solved in future React versions.
 				result.current.setData( {
 					intro: 'Hello World!',
 					content: ''
 				} );
 
-				await timeout( 200 );
+				await timeout( 500 );
```
gorzelinski, line 945: It's not super elegant, but I understand the problems with React.

Mati365: Yep :[

gorzelinski (approved):

Issue #567, opened 2025-01-06
data prop is not available for ckeditor


MohannedAbuzant:
Code:

```
import {ClassicEditor, EventInfo} from 'ckeditor5';
import React, {ReactNode} from 'react';
import {CKEditor} from '@ckeditor/ckeditor5-react';
import 'ckeditor5/ckeditor5.css';


function EditorComponent({
	value,
	schema,
	onChange
}: {
	value: string;
	schema: IEditor;
	onChange?: React.Dispatch<React.SetStateAction<{value: string}>>;
}): ReactNode {
	
	const handleOnChange = (_: EventInfo, editor: ClassicEditor) => {
		if (onChange) {
			onChange({value: editor.getData()});
		}
	};

	return (
		<>
			{schema.label && (
				<Text className={classes.label}>
					{schema.label}
					{isRequired && (
						<Text c="var(--mantine-color-error)" span ms="xs">
							*
						</Text>
					)}
				</Text>
			)}

			<CKEditor
				onChange={handleOnChange}
				editor={ClassicEditor}
				config={{
					licenseKey: 'GPL',
					// extraPlugins: [uploadPlugin], //Will import again when need Media
					toolbar: schema?.toolbar
						? schema?.toolbar
						: EditorDefaultSchema,
					plugins: EditorDefaultPluginsSchema,
					...(locale === LANGUAGES?.ar?.abbreviation && {
						translations: coreTranslations,
						language: {
							ui: 'ar'
						}
					})
				}}
				data={value}
			></CKEditor>
			<Text c={LAYOUT_COLORS.DANGER_COLOR}>{error}</Text>
		</>
	);
}

export default JbEditorComponent;
```

Problem:
data is not assignable to type
IntrinsicAttributes & IntrinsicClassAttributes<CKEditor<ClassicEditor>> & Readonly<Props<ClassicEditor>>
Property data does not exist on type
IntrinsicAttributes & IntrinsicClassAttributes<CKEditor<ClassicEditor>> & Readonly<Props<ClassicEditor>>

Versions:
"@types/react": "19.0.2",
    "@types/react-dom": "19.0.2"
    "next": "15.1.3",
    "@ckeditor/ckeditor5-react": "^9.4.0",
    "ckeditor5": "^44.1.0",

jonaldinger:
I'm getting the same error after upgrading to React v19. Has anybody figured out a solution or workaround?

Witoso:
[...]

PR #571, opened 2025-02-04
Bump React 19 version, remove `prop-types` from main component.

Mati365:

Fix: `data` prop is no longer unavailable in Next.js related integrations. Closes https://github.com/ckeditor/ckeditor5-react/issues/567
Other: Remove `prop-types` from dependencies of package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "./package.json": "./package.json"
   },
   "dependencies": {
-    "prop-types": "^15.7.2",
     "@ckeditor/ckeditor5-integrations-common": "^2.2.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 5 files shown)


Witoso:
https://github.com/facebook/prop-types was archived by facebook and is not maintained.

Witoso:
Should we bump to 19? https://majors.nullvoxpopuli.com/q?minors=false&old=false&packages=react

Would this cause disruptions for us developing against 18?

Mati365:
@Witoso Nope, absence of this package does nothing. Runtime validation of the props is no longer relevant in TypeScript era. I kept `prop-types` as dev dependency because our linter tends to use it.

gorzelinski (approved):
Seems to be working fine now.

PR #671, opened 2026-04-17
[React] Run integration tests for current and LTS editor branches in CI

### LINKED ISSUE #670: [React] Run integration tests for both LTS and current editor versions in CI
### Feature description

Update the `ckeditor5-react` CI setup so the integration test suite is executed against both supported CKEditor 5 release tracks: `LTS` and `current`.

### Definition of Done

- CI runs the React integration tests against both the `LTS` and `current` editor versions.
- If both tracks resolve to the same major version, the duplicated test run is skipped.
- The new setup is enabled in the standard CI flow.

### Technical details

- Adjust the CircleCI configuration to support two editor-version variants for integration testing.
- Consider extracting integration tests from the current `validate_and_tests` job so only the integration part is duplicated, while linting, build, type checks, and other tests remain unchanged unless needed.

pomek:
### 🚀 Summary

* Split CI into `validate`, `tests_current`, and an `tests_integration` matrix that explicitly covers `latest` and supported `lts-v*` editor tags.
* Standardize local and CI configuration around `CKEDITOR_*`, restore `.env.example`, and use Vite `envPrefix` plus `import.meta.env` instead of custom env plumbing.
* Define the integration-only Vitest project explicitly, run legacy multiroot compatibility assertions only on the real LTS path, and broaden the integration test globs so new real-editor tests are picked up automatically.

---

### 📌 Related issues

* Closes #670

---

### 💡 Additional information

* Internal-only CI, test, and docs changes, so no changelog entry is needed.
* `tests_current` runs the default Vitest project with coverage, while the CI matrix runs `--project integration` for `latest` and each supported LTS tag.
* The non-`latest` integration jobs use `CKEDITOR_LICENSE_KEY` from CI and skip themselves when the selected LTS tag resolves to the same major version as `latest`.


demos/shared/hooks/useCollaborationCredentials.ts
```diff
@@ -6,23 +6,23 @@ import { useMemo } from 'react';
...
+		const CLOUD_SERVICE_UPLOAD_URL = __CKEDITOR_UPLOAD_URL__;
 
 		const MISSING = (
 			[
-				[ 'VITE_CKE_LICENSE_KEY', LICENSE_KEY ],
-				[ 'VITE_CKE_TOKEN_URL', TOKEN_URL ],
-				[ 'VITE_CKE_WEBSOCKET_URL', WEBSOCKET_URL ]
+				[ 'CKEDITOR_LICENSE_KEY', LICENSE_KEY ],
+				[ 'CKEDITOR_TOKEN_URL', TOKEN_URL ],
+				[ 'CKEDITOR_WEBSOCKET_URL', WEBSOCKET_URL ]
```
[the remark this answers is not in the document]
pomek: I'm not sure if this is something I should care about. Need help to decide.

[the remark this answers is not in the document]
Mati365: imo, Cursor is correct here. I implemented these checks specifically for the RTC demo, which requires a custom license key. Furthermore, using `import.meta.env` aligns better with the ESM-native approach.

tests/multiroot/useMultiRootEditor.legacy-features.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+	return {
+		...actual,
+		getInstalledCKBaseFeatures: () => ( {
+			...actual.getInstalledCKBaseFeatures(),
+			elementConfigAttachment: false,
+			rootsConfigEntry: false
+		} )
+	};
+} );
```
Mati365: I'd avoid mocking here. Let's just run these legacy tests against the legacy CKEditor version instead.

.env.example
```diff
@@ -1,25 +0,0 @@
...
-# your Cloud Services secret key. The URL is called by the editor on
-# every session start (and periodically refreshed).
-VITE_CKE_TOKEN_URL=
-
-# Cloud Services WebSocket URL.
-# Provided by CKEditor Cloud Services (or your self-hosted CS instance).
-VITE_CKE_WEBSOCKET_URL=
-
-# (Optional) Cloud Services file upload URL.
-# This is the URL of YOUR file upload endpoint that accepts files uploaded
```
Mati365, line 24: Let's keep `.env.example` to follow standard conventions; `.env.dist` is too easy to mistake for a build-related file. I also prefer to keep the comments, as keys like `VITE_CKE_WEBSOCKET_URL` aren't clear enough on their own.

package.json
```diff
@@ -100,6 +100,7 @@
     "preview:19": "concurrently --kill-others -n \"BUILD,SERVE\" -c \"bgBlue.bold,bgGreen.bold\" \"REACT_VERSION=19 vite build --watch\" \"REACT_VERSION=19 vite preview\"",
     "build": "vite build && tsc --emitDeclarationOnly",
     "test": "vitest run --coverage",
+    "test:integration": "TEST_SCOPE=integration vitest run",
```
Mati365: I'm not sure it's needed.

README.md
```diff
@@ -45,6 +48,12 @@ To test the editor integration against a set of automated tests, run the followi
 pnpm run test
 ```
 
+To run only the real-editor integration coverage used by the CI version matrix, use:
+
+```bash
+pnpm run test:integration
+```
+
```
Mati365: I don't think that's the case. Looking at the setup, it doesn't seem to be running the tests on both legacy and latest versions as intended.

vite.config.ts
```diff
@@ -4,117 +4,147 @@
...
+	'tests/issues/354-destroy-editor-inside-context.test.tsx',
+	'tests/multiroot/useMultiRootEditor.test.tsx'
+];
 
-export default defineConfig( {
-	plugins: [
-		react( { jsxRuntime: 'classic' } )
-	],
+export default defineConfig( ( { mode } ) => {
+	const env = loadEnv( mode, __dirname, '' );
```
Mati365: `loadEnv` is automatically performed in `import.meta.env` approach.

vite.config.ts
```diff
@@ -4,117 +4,147 @@
...
+		// https://vitest.dev/config/
+		test: {
+			restoreMocks: true,
+			clearMocks: true,
+			mockReset: true,
+			unstubEnvs: true,
+			unstubGlobals: true,
+			setupFiles: [ './vitest-setup.ts' ],
+			include: TEST_SCOPE === 'integration' ? INTEGRATION_TESTS : [ 'tests/**/*.test.[j|t]sx' ],
+			exclude: TEST_SCOPE === 'non-integration' ? INTEGRATION_TESTS : [],
```
Mati365: Do we have any `non-integration` tests in integration package?

Mati365 (changes requested):

README.md
pomek: Addressed in `18e722c`. CI is now split into `validate`, `tests_current`, and an `integration_tests` matrix over `latest` and `lts-v47`, so the latest integration run is explicit.

.env.example
pomek: Restored in `0eb24f8`. We are back to a commented `.env.example` template and the docs plus demo error message now point to it.

package.json
pomek: Agreed, removed in `a9d6959`. The extra `test:integration` script is gone; CI and docs call Vitest directly with `--project integration`.

vite.config.ts
pomek: Addressed in `25b2fca`. We now use `envPrefix: CKEDITOR_` and read `import.meta.env` directly in demos and test setup, so the manual `loadEnv` plumbing is gone.

demos/shared/hooks/useCollaborationCredentials.ts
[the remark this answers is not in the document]
pomek: Addressed in `25b2fca`. The RTC demo now reads `CKEDITOR_LICENSE_KEY` from `import.meta.env` again, so the missing-variable validation works as before. The `GPL` fallback remains only in `vitest-setup.ts` for automated tests.

tests/multiroot/useMultiRootEditor.legacy-features.test.tsx
pomek: Addressed in `b0ebddb`. The mock is gone. `tests/multiroot/useMultiRootEditor.legacy-features.test.tsx` now feature-detects the installed CKEditor version and is included only in the integration project, so the real legacy path is exercised on the LTS matrix.

vite.config.ts
pomek: Addressed in `a9d6959`. The old `non-integration` branch is gone. Test selection now uses named Vitest projects: `default` and `integration`.

src/EditorWatchdogAdapter.tsx
```diff
@@ -63,6 +63,7 @@ export class EditorWatchdogAdapter<TEditor extends Editor> {
 		/* istanbul ignore else -- @preserve */
 		if ( config ) {
 			// <= 47 legacy config approach to watchdog configuration.
+			/* istanbul ignore next -- compatibility branch for older CKEditor 5 versions */
```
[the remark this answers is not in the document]
pomek: Addressed in `d9c62e6`. The contradictory Istanbul directives are gone from `EditorWatchdogAdapter`; both branches are covered by tests now, so the adapter no longer needs coverage hints there.

vite.config.ts
```diff
@@ -9,15 +9,32 @@ import { webdriverio } from '@vitest/browser-webdriverio';
...
+const DEFAULT_TESTS = [ 'tests/**/*.test.[j|t]sx' ];
+
+const INTEGRATION_TESTS = [
+	'tests/integrations/**/*.test.tsx',
+	'tests/issues/39-frozen-browser.test.tsx',
+	'tests/issues/349-destroy-context-and-editor.test.tsx',
+	'tests/issues/354-destroy-editor-inside-context.test.tsx',
+	'tests/multiroot/useMultiRootEditor.legacy-features.test.tsx',
+	'tests/multiroot/useMultiRootEditor.test.tsx'
+];
```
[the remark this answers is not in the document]
pomek: Addressed in `d9c62e6`. The integration project now uses directory globs for `tests/issues/*.test.tsx` and `tests/multiroot/*.test.tsx`, so new real-editor tests in those areas are picked up automatically instead of relying on a fully hardcoded file list.

.circleci/config.yml
```diff
@@ -149,6 +156,50 @@ jobs:
...
+            echo "Resolved ckeditor5@${EDITOR_VERSION_TAG} -> ${SELECTED_VERSION}"
+            echo "Resolved ckeditor5@latest -> ${LATEST_VERSION}"
+
+            if [[ "${EDITOR_VERSION_TAG}" != "latest" && "${SELECTED_VERSION%%.*}" == "${LATEST_VERSION%%.*}" ]];
+            then
+              echo "Skipping duplicated integration run because ${EDITOR_VERSION_TAG} and latest resolve to the same major version."
+              circleci-agent step halt
+            fi
+      - run:
+          name: Verify whether the selected track needs a hidden license key
```
pomek:
```suggestion
          name: Verify license key
```

In this case, _shorter_ means _better_.

.circleci/config.yml
```diff
@@ -149,6 +156,50 @@ jobs:
...
+      - image: cimg/node:24.11.0-browsers
+    parameters:
+      ckeditorDistTag:
+        type: string
+    steps:
+      - checkout
+      - bootstrap_repository_command
+      - browser-tools/install_chrome
+      - run:
+          name: Resolve editor versions and skip duplicated major
```
pomek:
```suggestion
          name: Resolve and deduplicate versions
```

src/multiroot/useMultiRootEditor.tsx
```diff
@@ -552,24 +558,23 @@ export const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookRe
...
-						attrs = {
-							...attrs,
+						} :
+						{
+							...baseAttrs,
 							data: rootData,
 							attributes: rootAttributes
 						};
-					}
+					/* istanbul ignore end -- compatibility branch for older CKEditor 5 versions */
```
pomek, line 576: I am still trying to figure out if these changes are needed. While I understand the `/* istanbul ...` comments, I am unsure about adding the ternary operator.

vite.config.ts
```diff
@@ -117,4 +147,5 @@ export default defineConfig( {
 		__REACT_VERSION__: REACT_VERSION,
 		__REACT_INTEGRATION_VERSION__: JSON.stringify( pkg.version )
 	}
-} );
+}
+);
```
pomek:
```suggestion
} );
```
Hm?

src/multiroot/useMultiRootEditor.tsx
Mati365: I don't think these changes are needed. I intentionally avoided too much ternary operators in order to make a little bit cleaner.

src/multiroot/useMultiRootEditor.tsx
```diff
@@ -435,13 +435,19 @@ export const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookRe
...
+					watchdog.setCreator( watchdogEditorCreator );
+					await watchdog.create( _getConfig() );
+				} :
+				async () => {
+					watchdog.setCreator( async ( _, config ) => watchdogEditorCreator( config ) );
+					await watchdog.create( data as any, _getConfig() );
+				};
+			/* istanbul ignore end -- compatibility branch for older CKEditor 5 versions */
+
+			await initializeWatchdog();
```
Mati365: I think that simple ` istanbul ignore start` and ` istanbul ignore end` would be enough here. I don't see any point in making it ternary.

src/multiroot/useMultiRootEditor.tsx
pomek: Applied in `48b5b04`. I replaced both compatibility ternaries with explicit `if/else` branches and kept the Istanbul guard around the compatibility-only block. The behavior and coverage stay the same, but the control flow is easier to read.

vite.config.ts
pomek: Applied in `48b5b04`. The file now closes with the standard `} );` form again.

src/multiroot/useMultiRootEditor.tsx
```diff
@@ -436,15 +436,19 @@ export const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookRe
 		try {
 			/* istanbul ignore if -- @preserve */
```
Mati365: That comment is no longer required

src/multiroot/useMultiRootEditor.tsx
```diff
@@ -435,13 +435,23 @@ export const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookRe
...
-				watchdog.setCreator( async ( _, config ) => watchdogEditorCreator( config ) );
-				await watchdog.create( data as any, _getConfig() );
+				initializeWatchdog = async () => {
+					watchdog.setCreator( async ( _, config ) => watchdogEditorCreator( config ) );
+					await watchdog.create( data as any, _getConfig() );
+				};
 			}
+			/* istanbul ignore end -- compatibility branch for older CKEditor 5 versions */
+
+			await initializeWatchdog();
```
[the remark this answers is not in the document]
pomek: Agreed and addressed in `dc32918`. The watchdog compatibility block is back to a direct `if/else`, which matches the adjacent attrs logic and removes the extra async indirection. I re-ran lint, the full default suite with coverage, the integration project, the build, and test type-checking; all passed.

.circleci/config.yml
```diff
@@ -149,6 +156,50 @@ jobs:
                 files: coverage/lcov.info
                 disable_search: true
 
+  integration_tests:
```
pomek:
```suggestion
  tests_integration:
```

This would match the pattern: `tests_{what}`.

vite-env.d.ts
```diff
@@ -3,15 +3,17 @@
...
+
 declare const __REACT_VERSION__: number;
 
 declare const __REACT_INTEGRATION_VERSION__: string;
 
 interface ImportMetaEnv {
-	readonly VITE_CKE_LICENSE_KEY: string;
-	readonly VITE_CKE_TOKEN_URL: string;
-	readonly VITE_CKE_WEBSOCKET_URL: string;
-	readonly VITE_CKE_UPLOAD_URL: string;
```
Mati365, line 14: Make sure it's working correctly. There is difference between `VITE_` prefixed envs and normal ones. https://vite.dev/guide/env-and-mode#env-variables

pomek: LGTM.

Image

Mati365 (dismissed):
LGTM

README.md
```diff
@@ -27,24 +27,34 @@ After cloning this repository, install necessary dependencies:
...
 ```bash
 pnpm run test
 ```
 
+To run only the real-editor integration suite used by the CI version matrix, use:
+
+```bash
+pnpm exec vitest run --project integration
+```
+
```
psmyrek, line 57: I think this section could be a bit more explicit about what this command actually covers locally.

The integration tests run against whichever `ckeditor5` and `ckeditor5-premium-features` versions are currently installed in `node_modules`, so by default is uses the `latest` version, not the LTS version. Reproducing the LTS version requires installing the matching `lts-v*` packages and setting `CKEDITOR_LICENSE_KEY`, like the CI job does.

It would help to document that local LTS reproduction needs the same steps as well, otherwise it is easy to assume that `pnpm exec vitest run --project integration` already covers the full matrix locally.

WDYT?

psmyrek (dismissed):
LGTM, with one minor comment. Feel free to skip it.

README.md
pomek: Updated the README to make this explicit. It now says that `pnpm exec vitest run --project integration` runs against whichever `ckeditor5` and `ckeditor5-premium-features` versions are currently installed in `node_modules`, and that reproducing the LTS lane locally requires installing the matching `lts-v*` packages and providing `CKEDITOR_LICENSE_KEY`, the same way the CI job does.

psmyrek (approved):

PR #675, opened 2026-05-18
Add support for paragraph like editor.

Mati365:
### 🚀 Summary

Add support for paragraph like editor.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9933
* Requires https://github.com/ckeditor/ckeditor5-integrations-common/pull/113

### DIFF
--- a/src/EditorElement.tsx
+++ b/src/EditorElement.tsx
@@ -3,15 +3,19 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-licensing-options
  */
 
-import React, { forwardRef, memo, type ElementType } from 'react';
+import React, { forwardRef, memo } from 'react';
 import { normalizeClassList } from './utils/normalizeClassList.js';
+import {
+	type EditorElementDefinition,
+	normalizeEditorElementDefinition
+} from './utils/normalizeEditorElementDefinition.js';
 
 /**
  * Creates and renders a dynamic React element based on the element definition owned and provided by the editor.
  *
  * @param props The component properties.
  */
-export const EditorElement = memo( forwardRef<HTMLElement, EditorElementProps>( ( { definition }, ref ) => {
+export const EditorElement = memo( forwardRef<HTMLElement, Props>( ( { definition }, ref ) => {
 	const { id, name: Tag, classes, styles, attributes } = normalizeEditorElementDefinition( definition ?? {
 		name: 'div'
 	} );
@@ -30,69 +34,9 @@ export const EditorElement = memo( forwardRef<HTMLElement, EditorElementProps>( 
 EditorElement.displayName = 'EditorElement';
 
 /**
- * Normalizes an editor element definition into a structured object.
- *
- * @param definition The definition to normalize.
- * @returns A strictly typed object definition containing at least the element name.
- */
-export function normalizeEditorElementDefinition( definition: EditorElementDefinition ): EditorElementObjectDefinition {
-	if ( typeof HTMLElement !== 'undefined' && definition instanceof HTMLElement ) {
-		throw new Error(
-			'An HTMLElement cannot be used as an editor element definition. ' +
-			'Please pass a string, a React component, or an object definition.'
-		);
-	}
-
-	if ( typeof definition !== 'object' || definition === null ) {
-		return {
-			name: definition
-		};
-	}
-
-	return definition as EditorElementObjectDefinition;
-}
-
-/**
- * Defines an editor element. It can be a basic React element type or a detailed configuration object.
- */
-export type EditorElementDefinition = ElementType | EditorElementObjectDefinition;
-
-/**
- * An object-based definition for an editor element, allowing customization
- * of classes, styles, and HTML attributes.
- */
-export type EditorElementObjectDefinition = {
-
-	/**
-     * The DOM tag name or React component to use.
-     */
-	name: ElementType;
-
-	/**
-     * The unique identifier (ID) to apply to the editable element.
-     */
-	id?: string;
-
-	/**
-     * Class name or array of class names to apply to the editable element. Each name can be provided as a string.
-     */
-	classes?: string | Array<string>;
-
-	/**
-     * Inline styles to apply to the editable element as a record of style properties.
-     */
-	styles?: Record<string, string>;
-
-	/**
-     * Additional DOM attributes to apply to the editable element.
-     */
-	attributes?: Record<string, string>;
-};
-
-/**
  * Properties for the {@link EditorElement} component.
  */
-export type EditorElementProps = {
+type Props = {
 
 	/**
      * The definition of the element to be rendered. Defaults to a `div` element if not provided or null.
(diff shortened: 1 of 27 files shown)


src/ckeditor.tsx
```diff
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-licensing-options
  */
 
-import React from 'react';
+import React, { type ElementType } from 'react';
```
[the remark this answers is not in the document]
Mati365: Expected

src/multiroot/EditorEditable.tsx
```diff
@@ -3,62 +3,66 @@
...
 
-		semaphore.runAfterMount( ( { instance } ) => {
-			if ( !innerRef.current ) {
-				return;
-			}
+		const options = root.getAttribute( '$rootEditableOptions' ) as RootEditableOptionsAttribute | undefined;
 
-			editor = instance;
+		return options ?? {};
+	}, [ root ] );
```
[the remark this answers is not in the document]
Mati365: It's not supposed to be updated after the root is created.

src/multiroot/useMultiRootEditor.tsx
```diff
@@ -627,26 +626,104 @@ export const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookRe
...
+		const { instance } = await semaphore.current!.waitFor();
+		const supports = getInstalledCKBaseFeatures();
+
+		instance.model.change( () => {
+			const mappedAttributes = {
+				...attributes,
+				...!supports.rootsConfigEntry && editableOptions && {
+					[ ROOT_EDITABLE_OPTIONS_ATTRIBUTE ]: editableOptions
+				}
+			};
```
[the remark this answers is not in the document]
Mati365: It's expected.

src/ckeditor.tsx
```diff
@@ -209,8 +213,14 @@ export default class CKEditor<TEditor extends Editor> extends React.Component<Pr
 	 * Render a <div> element which will be replaced by CKEditor.
 	 */
 	public override render(): React.ReactNode {
+		const { editor: Editor, config = {} } = this.props;
+		const definition = getEditorElementDefinition( Editor, config );
```
gorzelinski, line 217: I'm a bit worried about type safety here. In case of typo or passing a number, there won't be a proper error. We could narrow the type or reject more invalid shapes during normalization.

gorzelinski, line 217: Maybe we should add a comment that Classic creates its own editable inside the host wrapper, so the React-rendered element is just a placeholder.

src/multiroot/EditorEditable.tsx
```diff
@@ -3,64 +3,77 @@
...
-			editor = instance;
+		return { ...options };
+	}, [ root ] );
 
-			const { ui, model } = editor;
-			const root = model.document.getRoot( rootName );
+	useEffect( () => {
+		if ( !editor || !root || !rootEditableOptions ) {
+			return;
+		}
```
gorzelinski, line 38: As I see, options are either `null` (when root is `null`) or a truthy object. So, we can potentially simplify this, I guess.

```suggestion
	const rootEditableOptions = useMemo<RootEditableOptionsAttribute>( () => {
		const options = root?.getAttribute( ROOT_EDITABLE_OPTIONS_ATTRIBUTE ) as RootEditableOptionsAttribute | undefined;

		return options ?? {};
	}, [ root ] );

	useEffect( () => {
		if ( !editor || !root ) {
			return;
		}
```

tests/EditorElement.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+	} );
+
+	it( 'correctly forwards the ref to the rendered DOM element', () => {
+		const ref = createRef<HTMLElement>();
+
+		render( <EditorElement ref={ref} definition="span" /> );
+
+		expect( ref.current ).not.toBeNull();
+		expect( ref.current?.tagName ).toBe( 'SPAN' );
+	} );
```
gorzelinski, line 78: Maybe we should add a test about IDs, something like a 'should apply top-level `id` over any `attributes.id` in the definition.'

src/multiroot/useMultiRootEditor.tsx
```diff
@@ -627,37 +627,110 @@ export const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookRe
...
 
+	const removeRoot = async ( rootName: string ) => {
+		const { instance } = await semaphore.waitFor();
+
+		instance.model.change( () => {
+			instance.detachRoot( rootName, true );
+		} );
+	};
+
+	const addRoot = async (
```
gorzelinski, line 656: We should probably somehow handle errors in `addRoot`/`removeRoot`. Maybe `try/catch` inside?

src/EditorElement.tsx
```diff
@@ -0,0 +1,45 @@
...
+			{...attributes}
+			id={id}
+			className={normalizeClassList( classes )}
+			style={styles}
+		/>
+	);
+} ) );
+
+EditorElement.displayName = 'EditorElement';
+
```
gorzelinski, line 37: We forward those styles as `style={styles}`. React's style prop requires camelCase keys (marginTop, not margin-top). But the OSS-side normalises `$rootEditableOptions` for RTC and may persist whatever the user passes. Maybe worth either normalising at the boundary or documenting in this JSDoc.

gorzelinski (commented):
Besides those suggestions, I also found two occurrences of ckbox in version `2.9.2`. It's not the newest one. I'm unsure if that's intentional.

src/ckeditor.tsx
gorzelinski: That's probably a stale comment. It's about the code that now lives in the helper `getEditorElementDefinition()`.

src/EditorElement.tsx
Mati365: Good catch

src/multiroot/EditorEditable.tsx
Mati365: Nah, we can't as it's later used to pause react rendering until editor is loaded and root mounted.

src/multiroot/useMultiRootEditor.tsx
Mati365: I'd not do that. It'd be hard to debug why root is not added at all.

src/EditorElement.tsx
Mati365: Fixed in https://github.com/ckeditor/ckeditor5-react/pull/675/commits/cd4c0324bc4e2bb6579b446e4a12313b656b4c3d

src/ckeditor.tsx
Mati365: I don't think it's needed here. React will crash anyway with pretty straightforward message.

gorzelinski (approved):

PR #681, opened 2026-06-15
Add inline root support to RTC / Classic editor demos.

Mati365:
### 🚀 Summary

Add inline root support to RTC / Classic editor demos.

---

### 📌 Related issues

* Caused by https://github.com/ckeditor/ckeditor5-commercial/issues/10501


gorzelinski (approved):
I fixed nitpicks. Other than that - looks good.

PR #687, opened 2026-07-16
Upgrade ESLint to v10

martnpaneq:
Part of https://github.com/ckeditor/ckeditor5-internal/issues/4327.

Upgrades ESLint from v9 to **v10** and aligns the shared CKEditor 5 linter packages:

- `eslint` → `^10.0.0`
- `eslint-config-ckeditor5` → `^19.0.0` (where used)
- `eslint-plugin-ckeditor5-rules` → `^19.0.0` (where used)


przemyslaw-zan (changes requested):
- **`@eslint-react/dom-no-render: 'off'` is set repo-wide** (low). The legitimate need is the legacy `ReactDOM.render` in `demos/**` and `tests/**` (React 16/17); disabling it globally also silences it in `src/**`, so an accidental `ReactDOM.render` in production source wouldn't be flagged. I suggest scoping it to `demos/**` + `tests/**`, the way `static-components` was scoped to one file.
- Description nit: body says config `→ ^18.0.0`, diff ships `^19.0.0`.

przemyslaw-zan (approved):

PR #698, opened 2026-09-03
Apply the ratified dependency policy.

martnpaneq:
### 🚀 Summary

Applies the ratified dependency policy in this repository:

* Every automated install now says `--frozen-lockfile`, instead of relying on pnpm switching it on by itself once it notices CI.
* The `*ckeditor5*` cooldown exclusion is replaced by the five unscoped names we own. The glob matched *any* package with `ckeditor5` in the name, including someone else's, and an excluded name skips the cooldown completely.
* The remaining global installs name exact versions.
* A CODEOWNERS file is added - this repository had none, so nothing required a platform-team review of its lockfile or CI. It covers the dependency and CI paths the policy names, not the source directories.
* `blockExoticSubdeps: true` is added, so only registry tarballs can enter the tree as subdependencies.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4697.
* Policy and per-repository findings: https://github.com/ckeditor/ckeditor5-internal/issues/4665.
* `blockExoticSubdeps`: https://github.com/ckeditor/ckeditor5/issues/20216.

---

### 💡 Additional information

Checked with `pnpm install --frozen-lockfile --lockfile-only` on pnpm 11.14.0: the lockfile passes (956 entries), so `blockExoticSubdeps` does not reject anything in the tree today. The CircleCI jobs themselves were not run, so CI here is the real check.

Internal only (CI and dependency settings), so no changelog entry.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ commands:
           command: sudo npm i -g pnpm@11.14.0
       - run:
           name: Install dependencies
-          command: pnpm install
+          command: pnpm install --frozen-lockfile
 
   install_ssh_keys_command:
     description: "Install SSH keys"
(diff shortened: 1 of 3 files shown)


pomek (approved):
LGTM.

Issue #197, opened 2020-11-17
Getting "UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'destroy' of null" when testing a component that uses CKEditor


davesag:
We are using the CKEditor 5 React component in a project and having issues running unit tests.

Take this super paired back component

```js
import { CKEditor } from '@ckeditor/ckeditor5-react'
import ClassicEditor from '@ckeditor/ckeditor5-build-classic'

const TextEditor = ({ message, onReady, onChange, onBlur, onFocus }) => {
  return (
    <div style={{ width: '98vw' }}>
      <CKEditor
        editor={ClassicEditor}
        data={message}
        onReady={onReady}
        onChange={onChange}
        onBlur={onBlur}
        onFocus={onFocus}
      />
    </div>
  )
}
```

and Jest test

```js
import { render } from '@testing-library/react'

import TextEditor from '.'

describe('components > TextEditor', () => {
  const message = 'message'
  const onChange = jest.fn()
  const onFocus = jest.fn()
  const onReady = jest.fn()
  const onBlur = jest.fn()
  let component

  beforeAll(() => {
    component = render(
      <TextEditor
        message={message}
        onChange={onChange}
        onFocus={onFocus}
        onReady={onReady}
        onBlur={onBlur}
      />
    )
  })

  it('rendered without crashing', () => {
    expect(component.asFragment()).toMatchSnapshot()
  })
})
```

The test passes okay but something within the editor throws an error so we see this in the logs

```
(node:38593) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'destroy' of null
```

It's not a show-stopper, and we don't see the issue when actually running our app, but I'd love to know how to properly test the editor in a way that doesn't generate this warning.

pomek:
It looks like in some cases, the watchdog instance tries to destroy itself before its created and assigned properly.

https://github.com/ckeditor/ckeditor5-react/blob/f0a2e57bc507425ee9d038c55cda9db370425de1/src/ckeditor.jsx#L161-L164

cc: @ma2ciek

deniapps:
Got the same error on a different scenario, where we load CKEditor the second time via NextJS' [dynamic import](https://nextjs.org/docs/advanced-features/dynamic-import) 
It works fine with V2.1.0.

owboateng:
I am getting the same error. Please help fix it and it will be much appreciated.

yanlee26:
same here.

[...]

PR #699, opened 2026-09-11
Watchdog removal

mmotyczynska:
See: https://github.com/ckeditor/ckeditor5-commercial/issues/10827

### 🚀 Summary

*A brief summary of what this PR changes.*

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

*Optional: Notes on decisions, edge cases, or anything helpful for reviewers.*


mmotyczynska:
bugbot run

src/context/ckeditorcontext.tsx
```diff
@@ -17,77 +17,108 @@ import {
...
-	const [ currentContextWatchdog, setCurrentContextWatchdog ] = useState<ContextWatchdogValue<TContext>>( {
+	// Holds the context this component created. Destruction reads it instead of the state, because a
+	// context created moments before the component unmounts never makes it into a render.
+	const contextRef = useRef<TContext | null>( null );
+
+	// The state starts as 'initializing' because the CKEditor component checks it and waits for the context
+	// to be fully initialized before creating an editor in it.
+	const [ currentContext, setCurrentContext ] = useState<CKEditorContextValue<TContext>>( {
 		status: 'initializing'
 	} );
```
Mati365: I don't like the idea of having two sources of truth for obtaining an instance of context. I think it should be extracted into a separate hook. Something like:

```ts
const [ currentContext, setCurrentContext ] = useCKEditorContextState<TContext>( );
```

which has it's own ref that is being set after calling `setCurrentContext`.

src/context/ckeditorcontext.tsx
```diff
@@ -17,77 +17,108 @@ import {
...
 				status: 'initializing'
 			} );
 		}
+
+		return () => {
+			const context = contextRef.current;
+
+			contextRef.current = null;
+			context?.destroy();
+		};
```
Mati365: It'll be no longer needed after https://github.com/ckeditor/ckeditor5-react/pull/699/changes#r4004544912

src/context/ckeditorcontext.tsx
```diff
@@ -17,77 +17,108 @@ import {
...
 
-	// Cleanup the context watchdog when the component is unmounted. Abort if the watchdog is not initialized.
-	useEffect( () => () => {
-		if ( currentContextWatchdog.status === 'initialized' ) {
-			currentContextWatchdog.watchdog.destroy();
+	// Read when an error arrives rather than closed over, because the subscription below is tied to the
+	// context rather than to the render. A callback replaced on a later render has to be the one that runs.
+	const onErrorRef = useRef( onError );
+
+	onErrorRef.current = onError;
```
Mati365: I'd use https://github.com/ckeditor/ckeditor5-react/blob/master/src/hooks/useRefSafeCallback.ts here.

src/multiroot/useMultiRootEditor.tsx
```diff
@@ -374,92 +374,39 @@ export const useMultiRootEditor = ( props: MultiRootHookProps ): MultiRootHookRe
...
+		// Off the editor class rather than imported — see the note in `ckeditor.tsx`.
+		offEditorErrorRef.current = props.editor.onEditorError( ( { error, source } ) => {
+			if ( source !== instance ) {
+				return;
+			}
+
+			const onError = props.onError || console.error;
+
+			onError( error, { phase: 'runtime' } );
+		} );
```
Mati365: It should be moved to separate hook with unmount method.

Mati365 (changes requested):

src/ckeditor.tsx
Mati365, line 262: This condition is no longer needed for >= 49.

src/context/ckeditorcontext.tsx
mmotyczynska: I'd rather keep them apart, because they answer different questions.

The state drives rendering and carries `status` and `error`, which children read to know whether the context is ready. The ref exists only so the cleanup can destroy a context that was created moments before the component unmounted.

That case is why deriving one from the other does not work here: the effect closes over the state from its last run (`[ id, isLayoutReady ]`), while the context is created later, inside the async `then`. A cleanup reading the state would not see it, and the context would leak. Adding `currentContext` to the dependencies instead would re-initialize the context on every state change.

src/context/ckeditorcontext.tsx
mmotyczynska: Same thread as above: the cleanup is the only place that can still reach a context created after the last render, so it keeps reading the ref.

src/context/ckeditorcontext.tsx
mmotyczynska: Done — replaced with `useRefSafeCallback`, named `reportError` to match the same thing in the Vue integration.

src/multiroot/useMultiRootEditor.tsx
mmotyczynska: The unsubscribe isn't tied to unmount — it lives in `_destroyEditor()`, next to the tick that lets `<CKEditorContext />` go first when both teardowns fire at once (#197). A hook with its own unmount method would either have to repeat that coordination or unsubscribe at a different point in the lifecycle than it does today, so I'd rather not move it here. (In Vue the same subscription did sit in a plain `onBeforeUnmount`, which is why it moved into a composable there.)

src/ckeditor.tsx
mmotyczynska: I'd rather wait with the rest. The same comparison drives `rootsConfigEntry`, so it's eleven conditions across the three integrations, and which editor versions these majors will ultimately support isn't settled yet (ckeditor5-commercial#11613). Until it is, the dead branches don't break anything.

src/context/ckeditorcontext.tsx
Mati365: That's not the point. I understand the issue being fixed, and it does indeed work. My concern is that the component suddenly has two sources of truth from which a context instance can be retrieved - namely, `currentContext` and `contextRef`. I would prefer that we avoid this situation and provide only one possible way to obtain a context reference.

I checked whether the solution involving ref encapsulation (and thus avoiding component scope pollution) is possible, and it seems it is.

```ts
function useCKEditorContextState<TContext extends Context>() {
	const pendingContextInstanceRef = useRef<TContext | null>( null );
	const [ currentContext, setCurrentContext ] = useState<CKEditorContextValue<TContext>>( {
		status: 'initializing'
	} );

	const safeSetState = ( state: CKEditorContextValue<TContext> ) => {
		const pendingContext = pendingContextInstanceRef.current;

		pendingContextInstanceRef.current = null;
		void pendingContext?.destroy();

		if ( state.status === 'initialized' ) {
			pendingContextInstanceRef.current = state.context;
		}

		setCurrentContext( state );
	};

	useEffect( () => () => {
		pendingContextInstanceRef.current?.destroy();
	}, [] );

	return [ currentContext, safeSetState ] as const;
}
```

src/context/ckeditorcontext.tsx
Mati365: The `safeSetState` in the snippet below destroys the context, so it’s no longer needed:

https://github.com/ckeditor/ckeditor5-react/pull/699/changes/BASE..367beb002c23eeeb097630e5ed927ee41552bca2#r4012940987

src/multiroot/useMultiRootEditor.tsx
Mati365: > A hook with its own unmount method would either have to repeat that coordination or unsubscribe at a different point in the lifecycle than it does today, so I'd rather not move it here

It doesn’t really matter. All it needs to do is mount before any editor is initialized and remain mounted for the rest of the component’s lifetime. In that case, we’ll have continuous listening, and even changing the editor reference won’t require unmounting and remounting the listener. The only thing worth keeping in mind is to use `useRefSafeCallback` to always have the freshest instance reference in the handler. In any case, I’d stick with the built-in mechanism for unmounting side effects and avoid storing the unmount callback in a ref - in this case, it’s not required, and it reduces reusability and readability.

Mati365 (changes requested):

src/context/ckeditorcontext.tsx
mmotyczynska: Oh, now I get it — you're after one way in, not one holder. That's the better framing, and your hook does it cleanly: the ref stays where it has to be, but nothing outside can reach past it.

Applied, with two small departures. `void` is off in this repo's config, so the destroy calls are bare. And I renamed things slightly — `liveContextRef`, since it holds the live context rather than a pending one, and `setContextState`, so the name doesn't read as a plain setter when it also releases the previous context.

src/context/ckeditorcontext.tsx
mmotyczynska: Gone with it — the hook releases the previous context, so the cleanup had nothing left to do.

src/multiroot/useMultiRootEditor.tsx
mmotyczynska: Oh, I see what you mean now — subscribe for the component rather than for the editor, and let the handler read whichever instance is current. That's simpler than what was there, and it drops the ref entirely.

Done: one registration next to `editorRefs`, released by the effect's own cleanup, with `useRefSafeCallback` reading both the current instance and the current `onError`. Replacing the editor no longer takes the subscription down and sets it up again.

Mati365 (approved):
Looks fine.

martnpaneq (approved):

